PR #2, opened 2023-01-20
Upgrade Indexable [ADAM-82]

moeyensj:
This PR adds a variety of upgrades to the Indexable class (which really needs to be renamed: how about `IndexedDataTable` or `IndexedDataStructures`?)

Why was it developed? In short, I wanted to be able to have a class that allowed the user to store different data structures such as astropy Time objects, numpy arrays, numpy masked arrays, and then be able to slice the class as one would an array and have the class correctly slice its members. However, the different supported data structures all have different interfaces to do that kind of slicing. This is an effort to generalize this desired behavior behind the functions of a class and most importantly make it user-friendly as a whole. It's difficult to show the utility of the class without showing how classes that inherit from this class work in practice. I'd encourage looking at the unit tests that have been added to get an idea of what the class is doing. In two follow-up PRs, I'll add the upgrades to the coordinates class and the orbits class with examples.

Additionally, this class supports the idea of an index in to its underlying member data structures. Iteration always occurs over the index *not* the rows of the underlying data structures. The index can be equivalent to the range of the length of the class's underlying members in which case `Indexable.__iter__` will iterate over each row (this is the default behavior when no index is set). Why is an index useful? Imagine if you have orbits (N) that you propagate to a series of times (M), if you set the index to the orbit ID you can then iterate over the grouped ephemerides in groups of slices of length (M). Super useful for things like THOR, etc...

### DIFF
--- a/adam_core/utils/indexable.py
+++ b/adam_core/utils/indexable.py
@@ -333,7 +333,9 @@ class Indexable:
 
         for k, v in copy_self.__dict__.items():
             if k != "_class_index":
-                if isinstance(v, (list, np.ndarray, np.ma.masked_array, Time, Indexable)):
+                if isinstance(
+                    v, (list, np.ndarray, np.ma.masked_array, Time, Indexable)
+                ):
                     copy_self.__dict__[k] = v[member_ind]
                 elif isinstance(v, UNSLICEABLE_DATA_STRUCTURES):
                     copy_self.__dict__[k] = v
@@ -376,6 +378,8 @@ class Indexable:
 
             else:
                 self.__dict__[k] = np.delete(v, np.s_[class_ind], axis=0)
+
+        self.set_index(self._class_index_attribute)
 
         return
 
--- a/adam_core/utils/indexable.py
+++ b/adam_core/utils/indexable.py
@@ -108,8 +108,9 @@ class Indexable:
 
     Indexable maintains two array indexes:
         - class: the externally facing index (ie. the index into the class itself)
-        - members: the index of the members of the class (ie. the index into
-            the data structures carried by the class)
+        - members mapping: the index of the members of the class (ie. the index into
+            the data structures carried by the class which can either an array of
+            integer indices or an array of slices)
 
 
     Class Index:
@@ -122,9 +123,6 @@ class Indexable:
     If, as above, the class to members mapping is sorted (monotonically increasing), then
     it is converted to an array of slices which allows for faster querying.
     [slice(0, 2), slice(2, 5), slice(5, 8)]
-
-    Members Index:
-    [0, 1, 2, 3, 4, 5, 6, 7, 8]
 
     For example, here orbits is a subclass of Indexable and has members that are
     `~numpy.ndarray`s, `~numpy.ma.core.MaskedArray`s, and `~astropy.time.core.Time`s.
@@ -194,8 +192,6 @@ class Indexable:
         -------
         member_length : int
             The length of the members of the class.
-        member_index : `~numpy.ndarray`
-            The index of the members of the class.
 
         Raises
         ------
@@ -213,9 +209,7 @@ class Indexable:
             raise ValueError("All sliceable members must have the same length.")
         member_length = member_lengths.pop()
 
-        member_index = np.arange(0, member_length, dtype=int)
-
-        return member_length, member_index
+        return member_length
 
     def set_index(self, index_values: Optional[Union[str, npt.ArrayLike]] = None):
         """
@@ -241,8 +235,6 @@ class Indexable:
             underlying arrays. None if the arrays are grouped and can be represented as slices
         self._class_index_attribute : str, None
             The attribute on which the index was set, if any. None if no attribute was used.
-        self._member_index : `~numpy.ndarray`
-            The index of the members of the class.
         self._member_length : int
             The length of the members of the class.
 
@@ -253,7 +245,7 @@ class Indexable:
                     If the values are not None, str, or np.ndarray.
         """
         # Check if all the members are valid: have the same length
-        self._member_length, self._member_index = self._check_member_validity()
+        self._member_length = self._check_member_validity()
 
         # Figure out the values that are going to be mapped to an index
         # If no index values are given then we set the class index
@@ -320,12 +312,14 @@ class Indexable:
         member_ind : slice or np.ndarray
             Slice into this class's members.
         """
+        # Create array of all possible indices in the class's members
+        member_indices = np.arange(0, self._member_length)
+
         # --- Integer Slice
         # If the index is an integer then we need to convert it to a slice so that
         # we do not change the dimensionality of the member arrays (or in the cases
         # of a 1D array we need to avoid returning just a single value)
         if isinstance(class_ind, int):
-
             ind = slice(class_ind, class_ind + 1)
 
         # --- Slices, Arrays, and Lists
@@ -341,8 +335,10 @@ class Indexable:
         if isinstance(ind, slice) and ind.start is not None and ind.start >= len(self):
             raise IndexError(f"Index {ind.start} is out of bounds.")
 
-        elif isinstance(ind, slice) and self._class_index_to_slices is not None:
-
+        elif (
+            isinstance(ind, (slice, np.ndarray, list))
+            and self._class_index_to_slices is not None
+        ):
             slices = self._class_index_to_slices[ind]
 
             # If there is only one slice then we can just return it
@@ -367,16 +363,16 @@ class Indexable:
                     "Slices are not consecutive. "
                     "Combining slices and concatenating the members index for each slice."
                 )
-                return np.concatenate([self._member_index[s] for s in slices])
+                return np.concatenate([member_indices[s] for s in slices])
 
-        elif np.array_equal(self._class_index, self._member_index):
+        elif np.array_equal(self._class_index, member_indices):
             logger.debug("Using class index to index member arrays.")
             return self._class_index[ind]
         else:
             logger.debug(
                 "Using unique class index to index member arrays with np.isin."
             )
-            return self._member_index[
+            return member_indices[
                 np.isin(
                     self._class_index_to_integers,
                     self._class_index[ind],
--- a/adam_core/utils/indexable.py
+++ b/adam_core/utils/indexable.py
@@ -12,7 +12,7 @@ __all__ = ["Indexable", "concatenate"]
 
 logger = logging.getLogger(__name__)
 
-SLICEABLE_DATA_STRUCTURES = (np.ndarray, np.ma.masked_array, list, Time)
+SLICEABLE_DATA_STRUCTURES = (np.ndarray, np.ma.masked_array, Time)
 UNSLICEABLE_DATA_STRUCTURES = (str, int, float, dict, bool, set, OrderedDict)
 
 
@@ -333,7 +333,7 @@ class Indexable:
         for k, v in copy_self.__dict__.items():
             if k != "_class_index":
                 if isinstance(
-                    v, (list, np.ndarray, np.ma.masked_array, Time, Indexable)
+                    v, (np.ndarray, np.ma.masked_array, Time, Indexable)
                 ):
                     copy_self.__dict__[k] = v[member_ind]
                 elif isinstance(v, UNSLICEABLE_DATA_STRUCTURES):
@@ -365,7 +365,7 @@ class Indexable:
                         scale=v.scale,
                         format="mjd",
                     )
-                elif isinstance(v, (list, Indexable)):
+                elif isinstance(v, (Indexable)):
                     del v[member_ind]
                 elif isinstance(v, UNSLICEABLE_DATA_STRUCTURES):
                     self.__dict__[k] = v
@@ -400,42 +400,6 @@ class Indexable:
     def yield_chunks(self, chunk_size):
         for c in range(0, len(self), chunk_size):
             yield self[c : c + chunk_size]
-
-    def append(self, other):
-        assert type(self) == type(other)
-
-        for k, v in other.__dict__.items():
-            self_v = self.__dict__[k]
-
-            if isinstance(v, np.ma.masked_array):
-                self.__dict__[k] = np.ma.concatenate([self_v, v])
-                self.__dict__[k].mask = np.concatenate([self_v.mask, v.mask])
-                self.__dict__[k].fill_value = self_v.fill_value
-
-            elif isinstance(v, np.ndarray):
-                self.__dict__[k] = np.concatenate([self_v, v])
-
-            elif isinstance(v, Time):
-                self.__dict__[k] = Time(
-                    np.concatenate([self_v.mjd, v.mjd]), scale=v.scale, format="mjd"
-                )
-
-            elif isinstance(v, (list, Indexable)):
-                self_v.append(v)
-
-            elif isinstance(v, pd.MultiIndex):
-                self.set_index(index=self._index_attribute)
-
-            elif isinstance(v, UNSLICEABLE_DATA_STRUCTURES):
-                assert v == self_v
-
-            elif v is None:
-                self.__dict__[k] = None
-
-            else:
-                err = f"{type(v)} are not supported."
-                raise NotImplementedError(err)
-        return
 
     def sort_values(
         self,
@@ -544,7 +508,7 @@ class Indexable:
 
 def concatenate(
     indexables: List[Indexable],
-) -> Indexable:
+) -> "Indexable":
     """
     Concatenate a list of Indexables.
 
@@ -573,7 +537,7 @@ def concatenate(
     time_scales = {}
     time_formats = {}
     for k, v in indexables[0].__dict__.items():
-        if isinstance(v, (np.ndarray, np.ma.masked_array, list, Indexable)):
+        if isinstance(v, (np.ndarray, np.ma.masked_array, Indexable)):
             data[k] = [deepcopy(v)]
         elif isinstance(v, Time):
             time_attributes.append(k)
@@ -592,7 +556,7 @@ def concatenate(
     for indexable_i in indexables[1:]:
         for k, v in indexable_i.__dict__.items():
             if (
-                isinstance(v, (np.ndarray, np.ma.masked_array, list, Indexable))
+                isinstance(v, (np.ndarray, np.ma.masked_array, Indexable))
                 and k not in time_attributes
             ):
                 data[k].append(v)
@@ -604,33 +568,23 @@ def concatenate(
             else:
                 pass
 
-    N = 0
     for k, v in data.items():
         if isinstance(v, list):
             if isinstance(v[0], np.ma.masked_array):
                 copy.__dict__[k] = np.ma.concatenate(v)
-                copy.__dict__[k].fill_value = np.NaN
-                N = len(copy.__dict__[k])
             elif isinstance(v[0], np.ndarray) and k not in time_attributes:
                 copy.__dict__[k] = np.concatenate(v)
-                N = len(copy.__dict__[k])
             elif isinstance(v[0], Indexable):
                 copy.__dict__[k] = concatenate(v)
-                N = len(copy.__dict__[k])
             elif k in time_attributes:
                 copy.__dict__[k] = Time(
                     np.concatenate(v), scale=time_scales[k], format="mjd"
                 )
-                N = len(copy.__dict__[k])
-        elif isinstance(v, UNSLICEABLE_DATA_STRUCTURES):
-            pass
-        else:
-            pass
 
-    if "_index" in copy.__dict__.keys():
-        if copy._index_attribute is not None:
-            copy.set_index(copy._index_attribute)
+    if "_class_index" in copy.__dict__.keys():
+        if copy._class_index_attribute is not None:
+            copy.set_index(copy._class_index_attribute)
         else:
-            copy.set_index(np.arange(0, N, 1))
+            copy.set_index()
 
     return copy
--- a/adam_core/utils/indexable.py
+++ b/adam_core/utils/indexable.py
@@ -75,6 +75,29 @@ def _map_values_to_integers(values: npt.ArrayLike) -> npt.NDArray[int]:
     return np.array(values_mapped["values_mapped"])
 
 
+def _check_slices_are_consecutive(slices: npt.NDArray[slice]) -> bool:
+    """
+    Checks if an array of slices are consecutive and share the same step.
+
+    Parameters
+    ----------
+    slices : np.ndarray[slice]
+        The slices to be checked.
+
+    Returns
+    -------
+    bool
+        True if the slices are consecutive and share the same step.
+    """
+    for i, s_i in enumerate(slices[:-1]):
+        if s_i.stop != slices[i + 1].start:
+            return False
+        if s_i.step is not None and (s_i.step != slices[i + 1].step):
+            return False
+
+    return True
+
+
 class Indexable:
     """
     Class that enables indexing and slicing of itself and its members.
@@ -292,7 +315,7 @@ class Indexable:
 
         Returns
         -------
-        member_ind : np.ndarray
+        member_ind : slice or np.ndarray
             Slice into this class's members.
         """
         # --- Integer Slice
@@ -318,32 +341,31 @@ class Indexable:
 
         elif isinstance(ind, slice) and self._class_index_to_members_is_slice:
 
+            slices = self._class_index_to_members[ind]
+
+            # If there is only one slice then we can just return it
+            if len(slices) == 1:
+                return slices[0]
+
             # Check if the array of slices are consecutive and share
             # the same step size. If so, create a single slice that
             # combines all of the slices.
-            slices = self._class_index_to_members[ind]
-            is_consecutive = True
-            for i, s_i in enumerate(slices[:-1]):
-                if s_i.stop != slices[i + 1].start:
-                    is_consecutive = False
-                    break
-                if s_i.step is not None and (s_i.step != slices[i + 1].step):
-                    is_consecutive = False
-                    break
-
-            if is_consecutive:
+            elif _check_slices_are_consecutive(slices):
                 logger.debug(
                     "Slices are consecutive and share the same step. "
                     f"Combining slices a single slice with start {slices[0].start}, "
                     f"end {slices[-1].stop} and step {slices[0].step}."
                 )
                 return slice(slices[0].start, slices[-1].stop, slices[0].step)
+
+            # If the slices are not consecutive then we need to
+            # combine them and return the indices of the members
             else:
                 logger.debug(
                     "Slices are not consecutive. "
-                    "Combining slices a concatenating the members index for each slice."
+                    "Combining slices and concatenating the members index for each slice."
                 )
-                return np.concatenate([self._class_index[s] for s in slices])
+                return np.concatenate([self._member_index[s] for s in slices])
 
         elif np.array_equal(self._class_index, self._member_index):
             logger.debug("Using class index to index member arrays.")
--- a/adam_core/utils/indexable.py
+++ b/adam_core/utils/indexable.py
@@ -285,7 +285,7 @@ class Indexable:
                 )
                 return np.concatenate([self._class_index[s] for s in slices])
 
-        elif np.all(self._class_index == self._member_index):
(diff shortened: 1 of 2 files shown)


adam_core/utils/indexable.py
```diff
@@ -1,174 +1,446 @@
...
+    the len(instance) would be 3.
+
+    Class To Members Mapping:
+    [0, 0, 0, 1, 1, 1, 2, 2, 2]
+    If, as above, the class to members mapping is sorted (monotonically increasing), then
+    it is converted to an array of slices which allows for faster querying.
+    [slice(0, 2), slice(2, 5), slice(5, 8)]
+
+    Members Index:
+    [0, 1, 2, 3, 4, 5, 6, 7, 8]
```
spenczar, line 45: I don't really know what this means.

Is class something like the first dimension - the number of columns? A more concrete example would help.

Calling it "class" for the "externally facing index" makes me think that a better name might be "container". "Class" makes me think of class-level (as opposed to instance-level) behavior, which this definitely has nothing to do with.

adam_core/utils/indexable.py
```diff
@@ -1,174 +1,446 @@
...
+        """
+        Set an index on the class for the given values or attribute name.
+
+        Parameters
+        ----------
+        values : str, np.ArrayLike
+            The values to be indexed. If a string is given then the values are taken from the
+            attribute of the same name. If an array is given then the values are taken from the
+            array itself. If None is given then the index is set to the range of length
+            of the class's sliceable members.
```
spenczar, line 170: This level of type-based overloading seems pretty extreme to me. Why are all three behaviors under one method, as opposed to
```
def index_by(self, attrib_name: str):
    ...
  
def reindex(self, ordering: npt.ArrayLike):
    ...

# give this a better name - I don't understand what it does.
def index_by_none(self):
    ...

File: adam_core/utils/indexable.py [hunk 3]
```diff
@@ -1,174 +1,446 @@
...
+        """
+        Externally facing index.
+        """
+        return self._class_index
 
-        elif isinstance(i, tuple):
-            ind = list(i)
+    @class_index.setter
+    def class_index(self, value):
+        raise ValueError("Cannot set class_index directly. Use set_index() instead.")
```
spenczar, line 61: Why doesn't this call `set_index` for the user?

adam_core/utils/indexable.py
```diff
@@ -1,174 +1,446 @@
...
 
-    def _handle_index(self, i: Union[int, slice, tuple, list, np.ndarray]):
-        if isinstance(i, int):
-            if i < 0:
-                _i = i + len(self)
-            else:
-                _i = i
-            ind = slice(_i, _i + 1)
+    @property
+    def class_index(self):
```
spenczar, line 53: This property is only really used as a getter. Why not `def get_class_index(self) -> npt.ArrayLike`? (I'm not sure what the type is, actually).

More to the point, `_class_index` seems to me to be an internal implementation detail. Should it be accessible in the public interface at all? Keeping it as a private variable with no accessors seems fine.

adam_core/utils/indexable.py
```diff
@@ -1,174 +1,446 @@
...
+        Length of the class members.
+        """
+        return self._member_length
+
+    @member_length.setter
+    def member_length(self, value):
+        raise ValueError("Cannot set member_length directly. Use set_index() instead.")
 
+    @member_length.deleter
+    def member_length(self):
```
spenczar, line 155: My comments above on the `class_index` property apply equally to `class_index_to_members`, `class_index_to_members_is_slice`, `class_index_attribute`, `member_index`, and `member_length`. I would delete all of those.

adam_core/utils/indexable.py
```diff
@@ -1,174 +1,446 @@
...
+            The index of the members of the class.
+        self._member_length : int
+            The length of the members of the class.
+
+
+        Raises
+        ------
+        ValueError: If all sliceable members do not have the same length.
+        """
+        # --- Part 1: Determine the validity of this instance's members
```
spenczar, line 192: Instead of labeling sections, use more functions. Seems like you could use a `def _check_member_validity(self)` method.

adam_core/utils/indexable.py
```diff
@@ -1,174 +1,446 @@
...
+
+        # Scan members and if they are of a type that is sliceable
+        # then store the length. All sliceable members should have the same length
+        # along their first axis.
+        member_lengths = []
+        member_attributes = []
```
spenczar, line 198: `member_attributes` is unused, delete it

spenczar (changes requested):
Partial review - still reading set_index, but here's some comments early

spenczar (changes requested):
Ok, rest of my comments here.

I found this to be a challenging read. I still don't think that Indexable is very justified. It's a pretty abstract class with a lot going on, and a lot of performance consequences that seem hard to predict. It's quite generic which makes it hard to read.

I think that using dedicated container classes where we need them would make more sense than trying to write our own. We've been talking about that on slack - let's follow up there.

adam_core/utils/indexable.py
moeyensj: Fixed.

adam_core/utils/indexable.py
moeyensj: Fixed.

adam_core/utils/indexable.py
moeyensj: Done.

adam_core/utils/indexable.py
moeyensj: Done.

adam_core/utils/indexable.py
```diff
@@ -1,174 +1,446 @@
...
-                )
-            elif isinstance(v, (list, Indexable)):
-                del v[ind]
-            elif isinstance(v, pd.MultiIndex):
-                self.__dict__[k] = v.delete(ind)
-            elif isinstance(v, UNSLICEABLE_DATA_STRUCTURES):
-                self.__dict__[k] = v
-            elif v is None:
-                pass
+            if isinstance(v, SLICEABLE_DATA_STRUCTURES):
```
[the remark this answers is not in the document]
moeyensj: I'm going to spin out the discussion we had in Slack, which touches on this comment, into a Jira ticket / GitHub issue for us to address in the future.

adam_core/utils/indexable.py
```diff
@@ -1,174 +1,446 @@
...
+            and class_index_values.dtype.type != int
+        ):
+            logger.debug("Mapping class index values to integers.")
+            # We use pandas since numpy unique is slower for object dtypes
+            df = pd.DataFrame({"class_index_values": class_index_values})
+            df_unique = df.drop_duplicates(keep="first").copy()
+            df_unique["class_index_values_mapped"] = np.arange(0, len(df_unique))
+            class_index_values = df.merge(
+                df_unique, on="class_index_values", how="left"
+            )["class_index_values_mapped"].values
```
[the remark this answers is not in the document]
moeyensj: Done, I've split out the three logical chunks into functions, and made set_index calls them. I think it is a lot more readable.

adam_core/utils/tests/test_indexable.py
```diff
@@ -0,0 +1,393 @@
...
+        self.masked_array.mask = np.zeros(N)
+        self.masked_array.mask[0:N:2] = 1
+        self.times = Time(np.arange(59000, 59000 + N), scale="utc", format="mjd")
+
+        Indexable.__init__(self)
+
+
+def assert_equal(a, b):
+
+    if isinstance(a, (np.ndarray)) and isinstance(b, (np.ndarray)):
```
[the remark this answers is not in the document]
moeyensj: Rebased to fix this.

adam_core/utils/indexable.py
```diff
@@ -1,174 +1,446 @@
...
+
+        # --- Part 3: Now that we have the values that are going to be mapped to an index
+        # we can create the index.
+
+        # Extract unique values to act as the externally facing index.
+        self._class_index = pd.unique(class_index_values)
+
+        # If the class index values are monotonically increasing then we can convert the class index
+        # into an array of slices. This will make __getitem__ significantly faster.
+        is_sorted = np.all((class_index_values[1:] - class_index_values[:-1]) >= 0)
```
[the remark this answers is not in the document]
moeyensj: Fixed and pulled into a separate function.

adam_core/utils/indexable.py
```diff
@@ -1,174 +1,446 @@
...
+            ]
+        return member_ind
 
-            elif isinstance(v, (list, Indexable)):
-                self_v.append(v)
+    def __len__(self):
+        if self._class_index is not None:
+            return len(self._class_index)
+        else:
+            err = "Length is not defined for this class."
```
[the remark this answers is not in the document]
moeyensj: @spenczar Ah, great catch! So this was originally added to make sure that I remembered to initialize the Indexable class in its subclasses. As you point out, because the constructor calls `set_index` the length is always defined but if you subclass and forget the call to `super(self).__init__` then it won't work. Is this reasonable behavior? Should I modify the error to state that the Indexable was not properly initialized or just remove it and assume that the person writing the subclass knows to call the constructor?

adam_core/utils/indexable.py
```diff
@@ -1,174 +1,446 @@
...
+
+    def __delitem__(self, class_ind: Union[int, slice, tuple, list, np.ndarray]):
 
-            elif v is None:
-                self.__dict__[k] = None
+        member_ind = self.query_index(class_ind)
+
+        for k, v in self.__dict__.items():
+            # Everything but the class index is sliced as normal
+            if k != "_class_index":
```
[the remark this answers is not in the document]
moeyensj: Same as a previous comment: I'm going to spin out the discussion we had in Slack, which touches on this comment, into a Jira ticket / GitHub issue for us to address in the future.

adam_core/utils/indexable.py
```diff
@@ -1,174 +1,446 @@
...
+                    self._class_index_to_members,
+                    self._class_index[ind],
                 )
+            ]
+        return member_ind
 
-            elif isinstance(v, (list, Indexable)):
-                self_v.append(v)
+    def __len__(self):
+        if self._class_index is not None:
```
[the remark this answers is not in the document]
moeyensj: See the discussion in the other comment below.

adam_core/utils/indexable.py
```diff
@@ -1,174 +1,446 @@
...
         return
 
-    def __iter__(self):
-        for i in range(len(self)):
-            yield self[i]
+    def query_index(self, class_ind: Union[int, slice, list, np.ndarray]) -> np.ndarray:
+        """
+        Given a integer, slice, list, or `~numpy.ndarray`, appropriately slice
+        the class index and return the correct index or indices for this class's underlying
+        members.
```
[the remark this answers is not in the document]
moeyensj: Excellent points. `query_index` has been renamed to `_query_index`.

adam_core/utils/indexable.py
```diff
@@ -1,174 +1,446 @@
...
+            raise IndexError("class_ind should be either an int or a slice.")
 
-            elif isinstance(v, Time):
-                self.__dict__[k] = Time(
-                    np.concatenate([self_v.mjd, v.mjd]), scale=v.scale, format="mjd"
+        # --- Check boundaries on the slice
+        if isinstance(ind, slice) and ind.start is not None and ind.start >= len(self):
+            raise IndexError(f"Index {ind.start} is out of bounds.")
+
+        if isinstance(ind, int) and self._class_index_to_members_is_slice:
```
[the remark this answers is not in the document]
moeyensj: Removed. Good eyes!

adam_core/utils/indexable.py
moeyensj: I've replaced the property with just `index` that calls `set_index()` when the getter is called. In this case, it makes sense to keep this as the only public property.

adam_core/utils/indexable.py
```diff
@@ -1,174 +1,446 @@
...
+        Returns
+        -------
+        member_ind : np.ndarray
+            Slice into this class's members.
+        """
+        # --- Integer Slice
+        # If the index is an integer then we need to convert it to a slice so that
+        # we do not change the dimensionality of the member arrays (or in the cases
+        # of a 1D array we need to avoid returning just a single value)
+        if isinstance(class_ind, int):
```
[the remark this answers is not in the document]
moeyensj: Any class that inherits from Indexable would have a different interface if the underlying arrays change shape. It feels fairly natural to be able to select one orbit from an Orbits class that inherits from Indexable without needing to reshape the underlying arrays.

adam_core/utils/indexable.py
```diff
@@ -1,174 +1,446 @@
...
-                self.__dict__[k].mask = np.concatenate([self_v.mask, v.mask])
-                self.__dict__[k].fill_value = self_v.fill_value
+        # --- Slices, Arrays, and Lists
+        elif isinstance(class_ind, (slice, np.ndarray, list)):
+            ind = class_ind
 
-            elif isinstance(v, np.ndarray):
-                self.__dict__[k] = np.concatenate([self_v, v])
+        else:
+            raise IndexError("class_ind should be either an int or a slice.")
```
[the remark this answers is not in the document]
moeyensj: Fixed, was really just sloppy on my end.

adam_core/utils/indexable.py
```diff
@@ -1,174 +1,446 @@
...
+            logger.debug(
+                "Using unique class index to index member arrays with np.isin."
+            )
+            member_ind = self._member_index[
+                np.isin(
+                    self._class_index_to_members,
+                    self._class_index[ind],
                 )
+            ]
+        return member_ind
```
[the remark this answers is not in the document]
moeyensj: Done!

adam_core/utils/indexable.py
moeyensj: Added a clearer example in the new docstring.

adam_core/utils/indexable.py
```diff
@@ -1,174 +1,446 @@
...
+        else:
+            err = "Length is not defined for this class."
+            raise NotImplementedError(err)
 
-            elif isinstance(v, pd.MultiIndex):
-                self.set_index(index=self._index_attribute)
+    def __getitem__(self, class_ind: Union[int, slice, list, np.ndarray]):
+
+        member_ind = self.query_index(class_ind)
+        copy_self = copy(self)
```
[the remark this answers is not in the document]
moeyensj: It is weird for sure. And I'm trying to think of a use case where I wouldn't want to access the original arrays rather than work on a copy of them.. and I can't think of an example. 

What do you think of this? 
```
    def __getitem__(self, class_ind: Union[int, slice, list, np.ndarray]):

        member_ind = self._query_index(class_ind)
        item = self.__class__()

        for k, v in self.__dict__.items():
            if k != "_class_index":
                if isinstance(v, (np.ndarray, np.ma.masked_array, Time, Indexable)):
                    item.__dict__[k] = v[member_ind]
                elif isinstance(v, UNSLICEABLE_DATA_STRUCTURES):
                    item.__dict__[k] = v
                elif v is None:
                    pass
                else:
                    err = f"{type(v)} are not supported."
                    raise NotImplementedError(err)
            else:
                item.__dict__[k] = v[np.s_[class_ind]]

        return item
```

adam_core/utils/indexable.py
```diff
@@ -1,174 +1,446 @@
...
+                class_index_values = index_values
+                self._class_index_attribute = None
+
             else:
-                err = f"{type(v)} are not supported."
-                raise NotImplementedError(err)
+                err = "values must be None, str, or numpy.ndarray"
+                raise ValueError(err)
+
+        # If the index is to be set on an attribute that has an object or float dtype
```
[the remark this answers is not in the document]
moeyensj: Modified in rebase.

adam_core/utils/indexable.py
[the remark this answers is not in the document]
spenczar: I'd just remove it. When subclassing, you should **always** call the parent class's `__init__` method unless you have an extremely good reason not to.

adam_core/utils/indexable.py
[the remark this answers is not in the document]
spenczar: That looks scary to me. I don't think you can get away with `self.__class__()`. The class's initializer could easily take arguments, and you don't know what arguments to pass in at that spot. I think `copy` is preferable over that.

This might be something to optimize once data is stored in a separate attribute rather than bundled into `__dict__`. Then we could control the data structure used, so copying (or granting a view on a subslice) could be made efficient.

adam_core/utils/indexable.py
```diff
@@ -1,174 +1,451 @@
...
+        """
+        Scans the class's sliceable members and raises an error if they are
+        not all of the same length.
+
+        Sets
+        ----
+        self._member_length : int
+            The length of the members of the class.
+        self._member_index : `~numpy.ndarray`
+            The index of the members of the class.
```
spenczar, line 175: Usually code is cleaner if it returns values rather than setting variables as a side-effect. It will be safer for concurrent use, easier to test, and easier to read. I'd recommend changing this to return the two values. It can still raise an error if internal state is messed up because two members have different sizes.

adam_core/utils/indexable.py
```diff
@@ -1,174 +1,451 @@
...
+        # Scan members and if they are of a type that is sliceable
+        # then store the length. All sliceable members should have the same length
+        # along their first axis.
+        member_lengths = []
+        for k, v in self.__dict__.items():
+            if isinstance(v, SLICEABLE_DATA_STRUCTURES):
+                member_lengths.append(len(v))
 
+        if len(np.unique(member_lengths)) != 1:
+            raise ValueError("All sliceable members must have the same length.")
```
spenczar, line 190: This looks like a job for the Python builtin container `set`! It has O(1) insertion and uniqueness checks. Of course, the number of members will always be tiny, but it's worth knowing about as a faster and more idiomatic way to keep track of unique values.

```suggestion
        member_lengths = {len(v) for v in self.__dict__.values() if isinstance(v, SLICEABLE_DATA_STRUCTURES)}

        if len(member_lengths) != 1:
            raise ValueError("All sliceable members must have the same length.")
        member_length = member_lengths.pop()
```

adam_core/utils/indexable.py
```diff
@@ -1,174 +1,451 @@
...
+        if index_values is None:
+            class_index_values = np.arange(0, self._member_length)
+            self._class_index_attribute = None
 
-        return copy
+        else:
+            # Set the values that are to be indexed as a class index
+            if isinstance(index_values, str):
+                class_index_values = getattr(self, index_values)
+                self._class_index_attribute = index_values
```
spenczar, line 246: Maybe make this an `elif` switch:

```
        if index_values is None:
            # Use the range of the member lengths
            class_index_values = np.arange(0, self._member_length)
            self._class_index_attribute = None
        elif isinstance(index_values, str):
            # Use the provided string as an attribute lookup
            class_index_values = getattr(self, index_values)
            self._class_index_attribute = index_values
        elif isinstance(index_values, np.ndarray):
            # Use the provided values directly
            class_index_values = index_values
            self._class_index_attribute = None
        else:
            raise ValueError("values must be None, str, or numpy.ndarray")
```

adam_core/utils/indexable.py
```diff
@@ -1,174 +1,451 @@
...
-            elif isinstance(v, UNSLICEABLE_DATA_STRUCTURES):
-                copy.__dict__[k] = v
-            elif v is None:
-                pass
-            else:
-                err = f"{type(v)} are not supported."
-                raise NotImplementedError(err)
+        Parameters
+        ----------
+        values : str, np.ArrayLike
```
spenczar, line 205: The name here is out of sync with the function signature.

adam_core/utils/indexable.py
```diff
@@ -1,174 +1,451 @@
...
-            elif v is None:
-                pass
             else:
-                err = f"{type(v)} are not supported."
-                raise NotImplementedError(err)
+                err = "values must be None, str, or numpy.ndarray"
+                raise ValueError(err)
+
+        # If the index is to be set using an array that has an object or float dtype
+        # then we map the unique values of the index to integers. This will make querying the
```
spenczar, line 257: More accurately, if the array has a non-integer dtype.

adam_core/utils/indexable.py
```diff
@@ -1,174 +1,451 @@
...
+                    f"Combining slices a single slice with start {slices[0].start}, "
+                    f"end {slices[-1].stop} and step {slices[0].step}."
                 )
+                return slice(slices[0].start, slices[-1].stop, slices[0].step)
+            else:
+                logger.debug(
+                    "Slices are not consecutive. "
+                    "Combining slices a concatenating the members index for each slice."
+                )
+                return np.concatenate([self._class_index[s] for s in slices])
```
spenczar, line 351: This is very tricky, subtle code. I'd recommend giving it a function and thorough unit tests. I'm not able to verify its correctness by reading.

The `break`s with a boolean status variable are a dead giveaway that the consecutivity check should be a separate function that returns True or False.

adam_core/utils/indexable.py
```diff
@@ -1,174 +1,451 @@
...
+            of the class's sliceable members.
+
+        Sets
+        ----
+        self._class_index : `~numpy.ndarray`
+            The externally facing index of the class.
+        self._class_index_to_members : `~numpy.ndarray`
+            The mapping from the externally facing index to the class members.
+        self._class_index_to_members_is_slice : bool
+            Whether the mapping from the externally facing index to the class members are slices.
```
spenczar, line 218: This `..._is_slice` attribute sticks out to me. It shows that `self._class_index_to_members` is one variable being reused for two totally different purposes. It can be either used to build a slice-based index to unique values, or an integer mapping to non-integer values for attribute-based lookups. 

Maybe those should just be two different attributes: `self._class_index_unique_slices`, and `self._class_index_int_mapping`.

Even if you don't split it into two variables, I think `_class_index_to_members` is a very confusing name, since the values it holds are _not_ used as any sort of mapping to members. They are used as indices when looking inside `self._class_index`.

adam_core/utils/indexable.py
```diff
@@ -1,174 +1,451 @@
...
+    t1 = t0 + np.arange(0, 10) * u.day
+    backend = PYOORB()
+    propagated_orbits = backend.propagate_orbits(orbits, t1)
+
+    # The underlying arrays are now 10 * 3 in length
+    len(propagated_orbits) == 30
+
+    # If we update the index to be on orbit_id
+    # then the length of the class is 3 even though each
+    # member is 10 * 3 in length
```
spenczar, line 137: My head is really spinning on this... the idea is that you get a deduplicated index to the unique values? Yeah, I think I see that as I read the rest of the code more carefully. It's definitely not how I normally think of data, which is basically as tables. This is some other data structure, I'm not sure what I'd call it. I guess a hashmap of arrays, sort of, or something.

Feel free to just resolve this musing - I'm only posting it to record my thought process because this is pretty mind-bending and probably should get more explanation in the docs.

adam_core/utils/tests/test_indexable.py
```diff
@@ -0,0 +1,470 @@
...
+    else:
+        assert a == b
+
+    return
+
+
+def test__convert_grouped_array_to_slices_raises():
+    # Test that the function raises a ValueError if the array is not grouped
+    # (sorted but sort order does not matter)
+    array = np.concatenate([np.arange(0, 5), np.arange(0, 5)])
```
spenczar, line 78: nit: I'd find it clearer to spell out the array instead of concatenating function calls. For example:
```suggestion
    array = np.array([0, 1, 0, 1])
```

Maybe that's just because I don't use numpy as much, but I find it more legible.

adam_core/utils/tests/test_indexable.py
```diff
@@ -0,0 +1,470 @@
...
+    # The order of the sorted array should not matter as long as all elements that
+    # are supposed to be in the same slice are grouped together
+    array = np.arange(0, 10)[::-1]
+    slices = _convert_grouped_array_to_slices(array)
+    np.testing.assert_equal(slices, desired)
+
+    # Test that the function correctly converts a sorted array to slices
+    # this time with duplicates in the array
+    array = np.concatenate([np.arange(0, 5), np.arange(0, 5)])
+    array.sort()
```
spenczar, line 114: If you agree with my prior comment on legibility, maybe make this:
```
array = np.array([0, 0, 1, 1, 2, 2, 3, 3, 4, 4, 5, 5])
```
Or, better,
```
array = np.array([0, 0, 1, 1])
```

adam_core/utils/tests/test_indexable.py
```diff
@@ -0,0 +1,470 @@
...
+    values = np.array(["a", "a", "b", "c", "d", "e", "e", "e"])
+    mapping = _map_values_to_integers(values)
+    desired = np.array([0, 0, 1, 2, 3, 4, 4, 4])
+    np.testing.assert_equal(mapping, desired)
+
+    # Test that the function correctly maps float values with duplicates to integers
+    values = np.concatenate([np.arange(0, 10, dtype=float) for i in range(2)])
+    mapping = _map_values_to_integers(values)
+    desired = np.concatenate([np.arange(0, 10), np.arange(0, 10)])
+    np.testing.assert_equal(mapping, desired)
```
spenczar, line 162: Does `_map_values_to_integers` work if the values are in a scrambled order?

```
values = np.array(["a", "b", "a"])
mapping = _map_values_to_integers(values)
desired = np.array([0, 1, 0]) # maybe???
np.testing.assert_equal(mapping, desired)
```

spenczar (commented):
Way improved. The code is much better structured.

spenczar (approved):

adam_core/utils/tests/test_indexable.py
moeyensj: Yep -- that's actually what this test is testing! I'll add your exact example as an additional test for extra clarity.

adam_core/utils/indexable.py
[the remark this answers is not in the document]
moeyensj: Understood, I've left it as a copy.

adam_core/utils/indexable.py
moeyensj: Fixed.

adam_core/utils/indexable.py
moeyensj: Done.

adam_core/utils/indexable.py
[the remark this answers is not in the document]
moeyensj: Removed!

adam_core/utils/indexable.py
moeyensj: Fixed.

adam_core/utils/tests/test_indexable.py
moeyensj: Done!

adam_core/utils/indexable.py
moeyensj: Done.

adam_core/utils/indexable.py
moeyensj: Done!

adam_core/utils/indexable.py
moeyensj: Done, I've split it into two attributes. We can change the names in a future PR, but I have:
`class_index_to_slices` and `class_index_to_integers`. 

I'm actually thinking we can drop `member_index` at some point since that is always just an array of integers ranging from 0 to the length of the underlying members.

moeyensj: I've removed the member_index altogether. Instead, Indexable._query_index will compute the full indices into the member arrays and then slice them according to the indices passed to the function.

adam_core/utils/indexable.py
moeyensj: Great point. I've extracted the consecutive slice check into a separate function `_check_slices_are_consecutive` which will return a boolean (added some unit tests too).

adam_core/utils/indexable.py
spenczar: Oh, that's fantastic! Getting rid of `member_index` clarifies the code a lot. It seems that `_class_index` could just be renamed `_index` at this point, right?

moeyensj: Yeah actually, let me go ahead a submit a follow-up PR with some additional clean up.

PR #3, opened 2023-01-26
Add pytest-cov as a tests dependency, move testing to workflow

moeyensj:
Relatively simple PR: adds pytest-cov as a tests dependency and then drops the pytest command in the Docker recipe in favor of running tests as part of the GitHub workflow. 

These commits were originally in #2 but have been extracted here.


akoumjian (approved):
Looks good.

PR #5, opened 2023-02-02
Rename Indexable._class_index to Indexable._index

moeyensj:
As pointed out by @spenczar in a comment on #2, now that the concept of a members index is gone, we can rename the class index to just index. This makes the code a lot less of an eye sore. 

Renamed attributes  are: 
```
._class_index -> ._index
._class_index_to_slices -> ._index_to_slices
._class_index_to_integers -> ._index_to_integers
._class_index_attribute -> ._index_attribute
```

### DIFF
--- a/adam_core/utils/indexable.py
+++ b/adam_core/utils/indexable.py
@@ -113,16 +113,18 @@ class Indexable:
             integer indices or an array of slices)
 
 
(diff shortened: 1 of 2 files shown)


spenczar (approved):
Love it! Way easier to read

PR #8, opened 2023-02-15
Add Backend class [ADAM-107] 

moeyensj:
This PR is branched off of #7, so requires that to be merged first. 

This PR introduces a new "Backend" class that has several abstract methods that need to be defined by subclasses. Once implemented, the Backend class can then distribute those methods as multiple processes per chunk of orbits. 

One important change is that we make PYOORB the default integrator that comes with `adam_core` which I think is reasonable.  

Here is an example from the README:
```python
import numpy as np
from astropy.time import Time
from adam_core.orbits.query import query_horizons
from adam_core.backend import PYOORB

# Get orbit to propagate
initial_time = Time([60000.0], scale="tdb", format="mjd")
orbits = query_horizons(object_ids, initial_time)

# Make sure PYOORB is ready
backend = PYOORB()

# Define propagation times
times = initial_time + np.arange(0, 100) * u.d

# Propagate orbits! This function supports multiprocessing for large
# propagation jobs. 
propagated_orbits = backend.propagate_orbits(
    orbits,
    times,
    chunk_size=100,
    num_jobs=1,
)
```

Additional action items to be addressed in future PRs:
- [ ] Add observers class 
  - [ ] Add spiceypy utility to get observer coordinates
  - [ ] Add code to manage SPICE "kernel" files
- [ ] Add ephemeris class
- [ ] Add PYOORB ephemeris generation

### DIFF
--- a/.github/workflows/docker-build-lint-test.yml
+++ b/.github/workflows/docker-build-lint-test.yml
@@ -17,8 +17,12 @@ jobs:
       - uses: jpribyl/action-docker-layer-caching@v0.1.0
         # Ignore the failure of a step and avoid terminating the job.
         continue-on-error: true
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+        with:
+          install: true
       - name: Build and Install
-        run: docker build -t adam-core:$IMAGE_TAG .
+        run: docker build --load -t adam-core:$IMAGE_TAG .
       - name: Lint
         run:  docker run -i adam-core:$IMAGE_TAG pre-commit run --all-files
       - name: Test
--- a/.github/workflows/docker-build-lint-test.yml
+++ b/.github/workflows/docker-build-lint-test.yml
@@ -1,4 +1,4 @@
-name: docker - Build, Lint, and Test
+name: docker - Build Lint and Test
 
 on:
   push:
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM ubuntu:jammy
 
 RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
-    --mount=type=cache,target=/var/lib/apt,sharing=locked \ 
+    --mount=type=cache,target=/var/lib/apt,sharing=locked \
     apt-get update \
     && apt-get upgrade -y \
     && apt-get install libcurl4-openssl-dev libssl-dev git curl unzip python3.11 pip -y
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,26 @@
 FROM ubuntu:jammy
 
-RUN apt-get update \
+RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
+    --mount=type=cache,target=/var/lib/apt,sharing=locked \ 
+    apt-get update \
     && apt-get upgrade -y \
-    && apt-get install libcurl4-openssl-dev libssl-dev git python3.11 pip -y
+    && apt-get install libcurl4-openssl-dev libssl-dev git curl unzip python3.11 pip -y
+
+# Install pyoorb from B612's fork, which includes a patch to handle fortran errors better
+ENV OORB_TAG=v1.2.1a1.dev2
+ENV OORB_VERSION="pyoorb-1.2.1a1.dev2+66b7753.dirty"
+
+# Install oorb data
+RUN curl -fL -o /tmp/oorb_data.zip \
+    "https://github.com/B612-Asteroid-Institute/oorb/releases/download/${OORB_TAG}/oorb_data.zip"
+RUN unzip -d /opt/oorb_data /tmp/oorb_data.zip
+ENV OORB_DATA=/opt/oorb_data
+
+# Install pyoorb
+RUN --mount=type=cache,target=/root/.cache/pip \
+    export WHEEL_NAME="${OORB_VERSION}-cp310-cp310-manylinux_2_17_$(uname -m).manylinux2014_$(uname -m).whl" && \
+    pip install "https://github.com/B612-Asteroid-Institute/oorb/releases/download/${OORB_TAG}/${WHEEL_NAME}"
+
 
 # Upgrade pip to the latest version and install pre-commit
 RUN pip install --upgrade pip pre-commit
--- a/README.md
+++ b/README.md
@@ -127,6 +127,35 @@ orbits.cartesian
 orbits.spherical
 ```
 
+### Propagator
+The propagator class in `adam_core` provides a generalized interface to the supported orbit integrators and ephemeris generators. By default,
+`adam_core` ships with PYOORB.
+
+To propagate orbits with PYOORB (here we grab some orbits from Horizons first):
+```python
+import numpy as np
+from astropy.time import Time
+from astropy import units as u
+from adam_core.orbits.query import query_horizons
+from adam_core.propagator import PYOORB
+# Get orbit to propagate
+initial_time = Time([60000.0], scale="tdb", format="mjd")
+object_ids = ["Duende", "Eros", "Ceres"]
+orbits = query_horizons(object_ids, initial_time)
+# Make sure PYOORB is ready
+propagator = PYOORB()
+# Define propagation times
+times = initial_time + np.arange(0, 100) * u.d
+# Propagate orbits! This function supports multiprocessing for large
+# propagation jobs.
+propagated_orbits = propagator.propagate_orbits(
+    orbits,
+    times,
+    chunk_size=100,
+    num_jobs=1,
+)
+```
+
 ### Lower Level Classes
 These classes are designed more for developers interested in building codes derived from utilities
 in this package.
--- a/README.md
+++ b/README.md
@@ -12,84 +12,105 @@ To define an orbit:
 ```python
 from astropy.time import Time
 from adam_core.coordinates import KeplerianCoordinates
+from adam_core.coordinates import Times
+from adam_core.coordinates import Origin
 from adam_core.orbits import Orbits
 
-keplerian_elements = KeplerianCoordinates(
-    times=Time(59000.0, scale="tdb", format="mjd"),
-    a=1.0,
-    e=0.002,
-    i=10.,
-    raan=50.0,
-    ap=20.0,
-    M=30.0,
+keplerian_elements = KeplerianCoordinates.from_kwargs(
+    times=Times.from_astropy(
+        Time([59000.0], scale="tdb", format="mjd")
+    ),
+    a=[1.0],
+    e=[0.002],
+    i=[10.],
+    raan=[50.0],
+    ap=[20.0],
+    M=[30.0],
+    origin=Origin.from_kwargs(code=["SUN"]),
+    frame="ecliptic"
 )
-orbits = Orbits(
-    keplerian_elements,
-    orbit_ids="1",
-    object_ids="Test Orbit"
+orbits = Orbits.from_kwargs(
+    orbit_ids=["1"],
+    object_ids=["Test Orbit"],
+    coordinates=keplerian_elements.to_cartesian(),
 )
 ```
+Note that internally, all orbits are stored in Cartesian coordinates. Cartesian coordinates do not have any
+singularities and are thus more robust for numerical integration. Any orbital element conversions to Cartesian
+can be done on demand by calling `to_cartesian()` on the coordinates object.
+
 The underlying orbits class is 2 dimensional and can store elements and covariances for multiple orbits.
 
 ```python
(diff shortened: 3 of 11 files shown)


adam_core/backend/backend.py
```diff
@@ -0,0 +1,168 @@
...
+            p = mp.Pool(
+                processes=num_workers,
+            )
+
+            propagated_list = p.starmap(
+                propagation_worker,
+                zip(
+                    orbits_split,
+                    times_duplicated,
+                    backend_duplicated,
```
spenczar, line 78: Why does the backend need to be copied?

Also - probably worth loudly documenting that backends need to be picklable. `multiprocessing` uses `pickle` to send these things to the other processes. That means that backends can't, for example, have any open file handles or network connections.

adam_core/backend/pyoorb.py
```diff
@@ -0,0 +1,326 @@
...
+            pass
+        else:
+            if os.environ.get("OORB_DATA") is None:
+                os.environ["OORB_DATA"] = os.path.join(
+                    os.environ["CONDA_PREFIX"], "share/openorb"
+                )
+
+            # Prepare pyoorb
+            ephfile = os.path.join(os.getenv("OORB_DATA"), self.ephemeris_file)
+            err = oo.pyoorb.oorb_init(ephfile)
```
spenczar, line 49: Have you confirmed that this works with multiprocessing? `__init__` usually isn't run when unpickling a class instance, so I'd think that this `oorb_init` call would only happen in the main process and not the pool's subprocesses. I could be totally wrong though.

adam_core/backend/backend.py
```diff
@@ -0,0 +1,168 @@
...
+        the observers.
+
+        THIS FUNCTION SHOULD BE DEFINED BY THE USER.
+        """
+        pass
+
+    def generate_ephemeris(
+        self,
+        orbits: Orbits,
+        observers,
```
spenczar, line 107: Please annotate the type of `observers`

adam_core/backend/pyoorb.py
```diff
@@ -0,0 +1,326 @@
...
+        for k in kwargs:
+            if k not in allowed_kwargs:
+                warnings.warn(
+                    f"Invalid argument {k} passed to PYOORB backend. Ignoring..."
+                )
+
+        # If an allowed kwarg is missing, add the default
+        for k in allowed_kwargs:
+            if k not in kwargs:
+                kwargs[k] = PYOORB_CONFIG[k]
```
spenczar, line 33: Why not the more standard way of setting optional kwargs with defaults?

```suggestion
    def __init__(self, *, dynamical_model: str="N", ephemeris_file: str = "de430.dat"):
```

adam_core/backend/pyoorb.py
```diff
@@ -0,0 +1,326 @@
...
+
+        self.dynamical_model = kwargs["dynamical_model"]
+        self.ephemeris_file = kwargs["ephemeris_file"]
+
+        env_var = "ADAM_CORE_PYOORB"
+        if env_var in os.environ.keys() and os.environ[env_var] == "True":
+            pass
```
spenczar, line 40: What's the purpose behind this part?

adam_core/backend/pyoorb.py
```diff
@@ -0,0 +1,326 @@
...
+        else:
+            if os.environ.get("OORB_DATA") is None:
+                os.environ["OORB_DATA"] = os.path.join(
+                    os.environ["CONDA_PREFIX"], "share/openorb"
```
spenczar, line 44: Please check if CONDA_PREFIX is in env vars first

adam_core/backend/pyoorb.py
spenczar: Oh, I just read further and figured it out.

I still don't really understand, though! So this is to make sure you call oorb_init just once per process tree? Why?

I propose naming the env var something like `ADAM_CORE_PYOORB_INITIALIZED` to be a bit clearer.

adam_core/backend/pyoorb.py
```diff
@@ -0,0 +1,326 @@
...
+        Parameters
+        ----------
+        epochs : `~numpy.ndarray` (N)
+            Epoch in MJD to convert.
+        time_scale : {'UTC', 'UT1', 'TT', 'TAI'}
+            Time scale of the MJD epochs.
+
+        Returns
+        -------
+        epochs_pyoorb : `~numpy.ndarray (N, 2)
```
spenczar, line 210:
```suggestion
        epochs_pyoorb : `~numpy.ndarray` (N, 2)
```

adam_core/backend/pyoorb.py
```diff
@@ -0,0 +1,326 @@
...
+        if time_scale == "UTC":
+            time_scale_list = [1 for i in range(num_orbits)]
+        elif time_scale == "UT1":
+            time_scale_list = [2 for i in range(num_orbits)]
+        elif time_scale == "TT":
+            time_scale_list = [3 for i in range(num_orbits)]
+        elif time_scale == "TAI":
+            time_scale_list = [4 for i in range(num_orbits)]
+        else:
+            raise ValueError("time_scale should be one of {'UTC', 'UT1', 'TT', 'TAI'}")
```
spenczar, line 147: I think it's nice to use [`enum`](https://docs.python.org/3/library/enum.html) for things like this. You'd get type checking and the code will end up being a lot less verbose. 

```python
class OpenOrbTimescale(enum.Enum):
    UTC = 1
    UT1 = 2
    TT = 3
    TAI = 4

...

def _configure_orbits(
        orbits: np.ndarray,
        t0: np.ndarray,
        orbit_type: str,
        time_scale: OpenOrbTimescale,
        magnitude: Optional[float] = None,
        slope: Optional[float] = None,
    ) -> np.ndarray:

        ...
        time_scale_list = [time_scale.value for i in range(num_orbits)]   
```

Called like `PYOORB._configure_orbits(..., time_scale=OpenOrbTimescale.UTC, ...)`

adam_core/backend/pyoorb.py
```diff
@@ -0,0 +1,326 @@
...
+
+        if isinstance(slope, (float, int)):
+            slope_list = [slope for i in range(num_orbits)]
+        else:
+            slope_list = [0.15 for i in range(num_orbits)]
+
+        if isinstance(magnitude, (float, int)):
+            magnitude_list = [magnitude for i in range(num_orbits)]
+        else:
+            magnitude_list = [20.0 for i in range(num_orbits)]
```
spenczar, line 157: Looks like slope and magnitude are ignored if they are `ndarray`s, which doesn't seem right.

adam_core/backend/pyoorb.py
```diff
@@ -0,0 +1,326 @@
...
+                Omega : longitude of the ascending node in degrees
+                omega : argument of periapsis in degrees
+                M0 : mean anomaly in degrees
+            If cometary:
+                p : perihelion distance in AU
+                e : eccentricity in degrees
+                i : inclination in degrees
+                Omega : longitude of the ascending node in degrees
+                omega : argument of periapsis in degrees
+                T0 : time of perihelion passage in MJD
```
spenczar, line 97: What's the advantage of being able to pass in the keplerian and cometary representations to openorb, instead of always using cartesian?

adam_core/backend/pyoorb.py
```diff
@@ -0,0 +1,326 @@
...
+        orbit_ids_ = states[:, 0].astype(int)
+        x = states[:, 1]
+        y = states[:, 2]
+        z = states[:, 3]
+        vx = states[:, 4]
+        vy = states[:, 5]
+        vz = states[:, 6]
+        mjd_tt = states[:, 8]
+
+        # Convert output epochs to TDB
```
spenczar, line 292: This seems a bit surprising - why always convert to tdb?

adam_core/utils/multiprocessing.py
```diff
@@ -0,0 +1,36 @@
...
+    ----------
+    num_jobs : {"auto", int}
+        Number of jobs to launch.
+
+    Returns
+    -------
+    enable_parallel : bool
+        True if parallelization should be used (true for cases where num_jobs > 1).
+    num_workers : int
+        The number of workers to use.
```
spenczar, line 20: `enable_parallel` doesn't provide a lot more over returning just `num_workers`. Callers could check if `num_workers > 1`. I'd recommend trimming that down, I think it will feel simpler.

Also I'm not convinced that this adds much value in general, yeah. Callers could just use `mp.cpu_count()` themselves, this API is a bit indirect.

adam_core/backend/backend.py
```diff
@@ -0,0 +1,168 @@
...
+    propagated = backend._propagate_orbits(orbits, times)
+    return propagated
+
+
+def ephemeris_worker(orbits: Orbits, observers, backend: "Backend"):
+    ephemeris = backend._generate_ephemeris(orbits, observers)
+    return ephemeris
+
+
+class Backend(ABC):
```
spenczar, line 26: I think `Backend` is too generic of a name, and risks getting overloaded with tons of stuff. I'd recommend naming this `Propagator`. Ephemeris generation stands out as a little bit extra here (add an `EphemerisGenerator` ABC perhaps?) but I think that's acceptable enough.

spenczar (changes requested):
Agreed that `_check_parallel` can be discarded. Let callers set the number of workers to 1 or to `cpu_count()` themselves. You can disable using the multiprocessing pool when workers=1, I don't think the `_check_parallel` API is necessary for that.

The basic structure looks alright, but I think it could benefit from trimming down. Do we need to accept Keplerian orbital parameters in internal function calls, for example? I think shedding a lot of that weight would make this easier to read and maintain.

adam_core/backend/pyoorb.py
moeyensj: This is to match what JPL does: by default: orbits in TDB, observations in UTC

adam_core/backend/backend.py
moeyensj: I removed the copies and added calls to `itertools.repeat` rather than list comprehensions. Hopefully that is cleaner.

Also added a statement that subclasses must be pickleable.

adam_core/backend/pyoorb.py
moeyensj: If you call `oorb_init` twice from the same process it will return an error code 1. So this is an attempt to avoid that.

moeyensj: Also, renamed the env var as you suggested.

adam_core/backend/pyoorb.py
moeyensj: Done!

adam_core/backend/pyoorb.py
moeyensj: Fixed.

adam_core/backend/pyoorb.py
moeyensj: I've added two enums (one for orbit types and the other for time scales).

adam_core/backend/pyoorb.py
moeyensj: Ah so this comes from the olden days before I decided to go full cartesian. That said, we know openorb does not treat all representations equally so I've kept this in the event we ever want to try different representations. Happy to remove it if you want me to.

adam_core/backend/backend.py
moeyensj: Renamed to `Propagator`. I'd like to keep ephemeris generation as part of the same class that does propagation because you do both of those together so frequently.

adam_core/backend/pyoorb.py
moeyensj: Tested and confirmed that multiprocessing works!

adam_core/backend/pyoorb.py
moeyensj: Fixed. I actually re-did the openorb orbit formatting completely and now create a correctly formatted empty 2D  array rather which is then filled with the correct values. I think its miles better than this uglier format which was drawn from the pyoorb README.

adam_core/utils/multiprocessing.py
moeyensj: Removed, turns out the `mp.Pool` accepts both an int and None. If its None then it will use `mp.cpu_count()` processes.

adam_core/backend/backend.py
moeyensj: I'm going to hold off on this until I've defined the Observer's class in a follow up PR.

spenczar (approved):
Way improved. Nice!

moeyensj:
This PR will be rebased and modified once #17 is merged.

PR #12, opened 2023-02-23
Define __eq__ for Indexable and add tests

moeyensj:
Define `__eq__` for Indexable. For each member, check if it exists in the comparison instance and if it does ensure they are equal. 

Also adds tuples as unsliceable structures and increases unit test coverage.

### DIFF
--- a/adam_core/utils/indexable.py
+++ b/adam_core/utils/indexable.py
@@ -13,7 +13,7 @@ __all__ = ["Indexable", "concatenate"]
 logger = logging.getLogger(__name__)
 
 SLICEABLE_DATA_STRUCTURES = (np.ndarray, np.ma.masked_array, Time)
-UNSLICEABLE_DATA_STRUCTURES = (str, int, float, dict, bool, set, OrderedDict)
+UNSLICEABLE_DATA_STRUCTURES = (str, int, float, dict, bool, set, tuple, OrderedDict)
 
 
 def _convert_grouped_array_to_slices(values: npt.ArrayLike) -> npt.NDArray[slice]:
--- a/adam_core/utils/indexable.py
+++ b/adam_core/utils/indexable.py
@@ -399,6 +399,43 @@ class Indexable:
     def __len__(self):
         return len(self._index)
 
+    def __eq__(self, other: "Indexable") -> bool:
+        # If the objects do not share the same attributes then they are not equal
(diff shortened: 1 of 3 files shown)


adam_core/utils/indexable.py
```diff
@@ -399,6 +399,39 @@ def _query_index(
     def __len__(self):
         return len(self._index)
 
+    def __eq__(self, other: "Indexable") -> bool:
+        for k, v in self.__dict__.items():
```
spenczar, line 403: What if `other` has a key which `self` lacks?

One simple suggestion would be to check if `set(self.__dict__.keys()) == set(other.__dict__.keys())`.

adam_core/utils/tests/test_indexable.py
```diff
@@ -230,6 +245,131 @@ def test__convert_to_array():
     np.testing.assert_equal(Indexable._convert_to_array(value), desired)
 
 
+def test_Indexable__eq__():
```
spenczar, line 248: It's not a very big deal, but personally I'd break this into many many littler tests. I think it's easier to read, and if you have failures, you'll get _all_ the failures instead of the first one only.

adam_core/utils/indexable.py
moeyensj: Excellent catch. Will update as you suggest.

adam_core/utils/tests/test_indexable.py
moeyensj: Done!

adam_core/utils/tests/test_indexable.py
```diff
@@ -230,6 +245,154 @@ def test__convert_to_array():
...
+    indexable_1.__dict__["test"] = 1
+    indexable_2 = DummyIndexable(np.array([0, 1, 2]))
+    indexable_2.__dict__["test"] = 1
+    indexable_3 = DummyIndexable(np.array([0, 1, 2]))
+
+    assert indexable_1 == indexable_2
+    assert indexable_1 != indexable_3
+
+
+def test_Indexable__eq__list():
```
spenczar, line 370:
```suggestion
def test_Indexable__eq__set():
```

moeyensj: I need a coffee!

PR #13, opened 2023-03-01
Write test showing bug in Indexable.__getitem__

spenczar:
These tests currently fail.

### DIFF
--- a/adam_core/utils/indexable.py
+++ b/adam_core/utils/indexable.py
@@ -52,7 +52,7 @@ def _convert_grouped_array_to_slices(values: npt.ArrayLike) -> npt.NDArray[slice
     return np.array(slices)
 
 
-def _map_values_to_integers(values: npt.ArrayLike) -> npt.NDArray[int]:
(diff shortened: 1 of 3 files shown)


adam_core/utils/indexable.py
```diff
@@ -313,52 +313,97 @@ def _convert_to_array(value):
...
+        IndexError: If the index is out of bounds.
+        """
         # --- Integer Slice
         # If the index is an integer then we need to convert it to a slice so that
         # we do not change the dimensionality of the member arrays (or in the cases
         # of a 1D array we need to avoid returning just a single value)
         if isinstance(class_ind, int):
-            ind = slice(class_ind, class_ind + 1)
+            # Check boundaries on the integer
+            if class_ind >= len(self):
```
spenczar, line 342: What about negative index values?

moeyensj: Great point, fixed with [fefaf45](https://github.com/B612-Asteroid-Institute/adam_core/pull/13/commits/fefaf45f8a02ae5b430d1a96b6cebfb5f4c90005).

PR #15, opened 2023-03-18
Benchmarks

spenczar:
Added some basic benchmarks. 

You can run them with `pytest --benchmark-only`. They're disabled by default.

My laptop's results:
image

### DIFF
--- a/adam_core/coordinates/tests/test_benchmarks.py
+++ b/adam_core/coordinates/tests/test_benchmarks.py
@@ -1,9 +1,8 @@
-import pytest
-import numpy as np
 import astropy.time
+import numpy as np
+import pytest
 
 from ..cartesian import CartesianCoordinates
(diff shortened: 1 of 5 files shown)


adam_core/coordinates/tests/test_benchmarks.py
```diff
@@ -0,0 +1,39 @@
...
+        vz=np.array([1]),
+        times=astropy.time.Time(50000, format="mjd"),
+    )
+    benchmark(
+        transform_coordinates,
+        coords=from_coords,
+        representation_out=representation,
+        frame_out=frame,
+        origin_out=origin,
+    )
```
moeyensj, line 39: I need to add some actual tests of the coordinate and covariance transformations.

moeyensj (approved):
This is an excellent contribution.

PR #16, opened 2023-04-24
Add pre-commit

moeyensj:
... and run it on all files.


adam_core/orbits/orbits.py
```diff
@@ -111,18 +111,16 @@ def to_df(
...
-            determine if any uncertainties are present and add them if so.
+            Include lower triangular covariance matrix columns.
 
         Returns
         -------
         df : `~pandas.DataFrame`
             Pandas DataFrame containing orbits.
         """
-        df = CoordinateMembers.to_df(
+        df = CoordinateMembers._to_df(
```
akoumjian, line 123: Why explicitly call the parent class here?

adam_core/orbits/orbits.py
```diff
@@ -169,7 +167,7 @@ def from_df(
         cls : `~adam_core.orbits.Orbits`
             Orbits class.
         """
-        data = cls._dict_from_df(
+        data = CoordinateMembers._dict_from_df(
```
akoumjian, line 170: Same here, is there something with the inheritance that I can't remember?

akoumjian (changes requested):
Approved, pending license update.

adam_core/orbits/orbits.py
moeyensj: At some point, there was some ambiguity as two parent classes had `.to_df` implemented but that has been fixed.

adam_core/orbits/orbits.py
moeyensj: Fixed.

moeyensj:
@akoumjian Re-licensed to MIT, I've kept the THOR attribution.

akoumjian (approved):

PR #21, opened 2023-06-27
Unit tests for orbital element transformations

moeyensj:
Adds tests for Keplerian, Cometary, and Cartesian orbital element transformations using data from JPL Horizons.

The tolerances with respect to JPL's conversions (via Horizons) are not as good as I would have expected. I don't fully understand why, I suspect that the way I'm using Horizons to generate the test data isn't actually doing what I intended. 

Instead, I've also added tests against SPICE's [oscelt](https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/FORTRAN/spicelib/oscelt.html) and [conics](https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/FORTRAN/spicelib/conics.html) functions as a third source to test against. **The agreement with SPICE's conversions is excellent**:
- Keplerian/Cometary to Cartesian: positions to within 10 mm (6.684e-14 au), velocities to within 10 nm/s (5.775e-15 au / d)
- Cartesian to Keplerian/Cometary: semi-major axis / periapse distance to within 10 mm, angles to within 10 nanoarcseconds, eccentricity to within 1e-15. 

The one untested coordinate dimension is time of periapse passage as I have not been able to find equivalent functionality in SPICE... I'll do some more digging.


adam_core/coordinates/tests/test_transforms_cometary.py
```diff
@@ -0,0 +1,174 @@
...
+    cartesian_elements_expected = orbital_elements[
+        ["x", "y", "z", "vx", "vy", "vz"]
+    ].values
+    t0_mjd = orbital_elements["mjd_tdb"].values
+
+    cartesian_elements_actual = cometary_to_cartesian(cometary_elements, t0_mjd)
+    diff = cartesian_elements_actual - cartesian_elements_expected
+
+    # Calculate offset in position in mm
+    r_diff = np.linalg.norm(diff[:, :3], axis=1) * u.au.to(u.mm)
```
akoumjian, line 26: I don't have a solution, because ndarray is likely the primitive we want to do most of this in, but I find the select by column index hard to follow and I worry about how brittle it is.

adam_core/coordinates/tests/test_transforms_keplerian.py
```diff
@@ -0,0 +1,230 @@
...
+
+    # Calculate offset in periapsis distance in m
+    q_diff = np.abs(diff[:, 1]) * u.au.to(u.m)
+    # Assert periapsis distance is to within 100 m
+    np.testing.assert_array_less(q_diff, 200)  # TODO
+
+    # Calculate offset in apoapsis distance in m
+    Q_diff = np.abs(diff[:, 2]) * u.au.to(u.m)
+    # Assert apoapsis distance is to within 100 m
+    np.testing.assert_array_less(Q_diff, 200)  # TODO
```
akoumjian, line 97: Is this what you were saying about disagreement with horizon? Currently looks like we are testing within 200m?

adam_core/coordinates/tests/test_transforms_spice.py
```diff
@@ -0,0 +1,242 @@
...
+def test_cartesian_to_keplerian_elliptical_against_spice(orbital_elements):
+    # Test cartesian_to_keplerian (vmapped) against the keplerian elements
+    # calculated by SPICE for a series of sample orbital elements.
+
+    # Limit to elliptical orbits
+    orbital_elements = orbital_elements[orbital_elements["e"] < 1]
+
+    t0 = orbital_elements["mjd_tdb"].values
+    cartesian_elements = orbital_elements[["x", "y", "z", "vx", "vy", "vz"]].values
+    t0_jd = t0 + 2400000.5
```
akoumjian, line 104: Not sure if it accounts for any of our disagreements, but floating point math here will stop working at a precision of about `0.0000000001`.

adam_core/dynamics/kepler.py
```diff
@@ -6,6 +6,7 @@
 from .barker import solve_barker
 
 config.update("jax_enable_x64", True)
+config.update("jax_platform_name", "cpu")
```
akoumjian, line 9: Can jax do automatic platform detection? Guessing we want to set this intentionally somewhere at runtime.

akoumjian (approved):
One additional set of tests might include measuring drift over repeated transformations back and forth (say, between keplerian and cartesian). I believe there are times when we need to move back and forth between cartesian space and it might be good to see what kind of limit we support before the drift starts to dramatically accumulate.

adam_core/dynamics/kepler.py
moeyensj: Yes, it does by default. This is just a simple workaround for it to stop complaining about not finding a TPU/GPU to run on. We can remove these statements once we start running on more than just CPUs.

PR #23, opened 2023-07-11
Add 2-body propagation

moeyensj:
Ports some of the dynamics modules and 2-body propagation code from THOR to adam_core. 

Our agreement with SPICE is excellent: 2-body propagator agrees with `spiceypy` 2-body propagations to within 10 cm in position and 1 mm/s in velocity

Here is a snippet from the README:

```python
import numpy as np
from astropy.time import Time
from adam_core.orbits.query import query_sbdb
from adam_core.dynamics.propagation import propagate_2body

# Get orbit to propagate
orbits = query_sbdb(object_ids)

# Define propagation times
times = Time(np.arange(59000, 60000), scale="tdb", format="mjd")

# Propagate orbits with 2-body dynamics
propagated_orbits = propagate_2body(
    orbits,
    times
)
```

Note that covariances are also propagated with `jax`. :)

### DIFF
--- a/adam_core/dynamics/chi.py
+++ b/adam_core/dynamics/chi.py
@@ -1,0 +1,134 @@
+from typing import Tuple
+
+import jax.numpy as jnp
+import numpy as np
+from jax import config, jit, lax
+
+from ..constants import Constants as c
+from .stumpff import calc_stumpff
+
+config.update("jax_enable_x64", True)
+
+MU = c.MU
+
+
+@jit
+def calc_chi(
+    r: jnp.ndarray,
+    v: jnp.ndarray,
+    dt: float,
+    mu: float = MU,
+    max_iter: int = 100,
+    tol: float = 1e-16,
+) -> Tuple[
+    np.float64, np.float64, np.float64, np.float64, np.float64, np.float64, np.float64
+]:
+    """
+    Calculate universal anomaly chi using Newton-Raphson.
+
+    Parameters
+    ----------
+    r : `~jax.numpy.ndarray` (3)
+        Position vector in au.
+    v : `~jax.numpy.ndarray` (3)
+        Velocity vector in au per day.
+    dt : float
+        Time from epoch to which calculate chi in units of decimal days.
+    mu : float
+        Gravitational parameter (GM) of the attracting body in units of
+        au**3 / d**2.
+    max_iter : int
+        Maximum number of iterations over which to converge. If number of iterations is
+        exceeded, will return the value of the universal anomaly at the last iteration.
+    tol : float
+        Numerical tolerance to which to compute chi using the Newtown-Raphson
+        method.
+
+    Returns
+    -------
+    chi : float
+        Universal anomaly.
+    c0, c1, c2, c3, c4, c5 : 6 x float
+        First six Stumpff functions.
+
+    References
+    ----------
+    [1] Curtis, H. D. (2014). Orbital Mechanics for Engineering Students. 3rd ed.,
+        Elsevier Ltd. ISBN-13: 978-0080977478
+    """
+    v_mag = jnp.linalg.norm(v)
+    r_mag = jnp.linalg.norm(r)
+    rv_mag = jnp.dot(r, v) / r_mag
+    sqrt_mu = jnp.sqrt(mu)
+
+    # Equations 3.48 and 3.50 in Curtis (2014) [1]
+    alpha = -(v_mag**2) / mu + 2 / r_mag
+
+    # Equation 3.66 in Curtis (2014) [1]
+    chi = sqrt_mu * jnp.abs(alpha) * dt
+
+    ratio = 1e15
+    iterations = 0
+    # Define parameters array (arguments that will be changing as
+    # the while loop iterates):
+    # chi, c0, c1, c2, c3, c4, c5, ratio, iterations
+    p = [chi, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, ratio, iterations]
+
+    # Define while loop body function
+    @jit
+    def _chi_newton_raphson(p):
+
+        chi = p[0]
+        ratio = p[-2]
+        iterations = p[-1]
+
+        chi2 = chi**2
+        psi = alpha * chi2
+        c0, c1, c2, c3, c4, c5 = calc_stumpff(psi)
+
+        # Newton-Raphson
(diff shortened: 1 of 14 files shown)


adam_core/dynamics/stumpff.py
```diff
@@ -0,0 +1,89 @@
...
+    c0 = jnp.cosh(sqrt_npsi)
+    c1 = jnp.sinh(sqrt_npsi) / sqrt_npsi
+
+    # Equation 6.9.16 in Danby (1992) [1]
+    # states the recursion relation for higher
+    # order Stumpff functions
+    c2 = (1.0 - c0) / psi
+    c3 = (1.0 - c1) / psi
+    c4 = (1 / 2.0 - c2) / psi
+    c5 = (1 / 6.0 - c3) / psi
```
spenczar, line 44: So many data dependencies here... it makes sense that this was the best in the 60s when Stumpff was working on the Apollo computer, but I bet someone really smart could do better for today's CPU architectures.  (obviously, not a concern for this PR :P)

adam_core/dynamics/propagation.py
```diff
@@ -0,0 +1,174 @@
...
+    n_times = len(times)
+    orbit_ids_ = np.repeat(orbit_ids, n_times)
+    object_ids_ = np.repeat(object_ids, n_times)
+    orbits_array_ = np.repeat(cartesian_orbits, n_times, axis=0)
+    t0_ = np.repeat(t0, n_times)
+    t1_ = np.tile(t1, n_orbits)
+
+    orbits_propagated = _propagate_2body_vmap(
+        orbits_array_, t0_, t1_, mu, max_iter, tol
+    )
```
spenczar, line 128: It strikes me that, for the same orbit, and the same t0, we worth through a sequence of `t1_` values - it's the only parameter that is varying. If you chase that through the code, the _only instruction_ that is different is https://github.com/B612-Asteroid-Institute/adam_core/blob/633f818f0242a89211a7ff12125fb1f7acb2d808/adam_core/dynamics/chi.py#L68. So, _in theory_, that is the first point where vectorization ought to occur; all the operations before there are identical across the orbit's time steps.

Could make the code marvelously complicated to accommodate that, though...

adam_core/dynamics/propagation.py
```diff
@@ -0,0 +1,174 @@
...
+    orbits_propagated = np.array(orbits_propagated)
+
+    cartesian_covariances = orbits.coordinates.covariance.to_matrix()
+    if not np.all(np.isnan(cartesian_covariances)):
```
spenczar, line 132: Trying to work this out - this is "if any of the cells in any of the covariance matrices is non-null", right?

Is `transform_covariances_jacobian` OK with _some_ nulls in a particular covariance matrix?

spenczar (approved):

adam_core/dynamics/chi.py
```diff
@@ -0,0 +1,134 @@
+from typing import Tuple
+
+import jax.numpy as jnp
+import numpy as np
+from jax import config, jit, lax
+
+from ..constants import Constants as c
+from .stumpff import calc_stumpff
+
+config.update("jax_enable_x64", True)
```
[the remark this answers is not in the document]
moeyensj: Excellent point and it's fairly easy to test: we just disable the config update and run our tests which include benchmarks and tests against `spiceypy` for accuracy.

adam_core/dynamics/propagation.py
moeyensj: I haven't tried but I suspect it will fail. We may need to approximate nulls in a covariance matrix as some very large number to give them the appearance of being unconstrained. In most cases, I suspect if we have a full state vector we will have either no covariance matrix or a fully defined one.

moeyensj: To clarify, if any cells in this 3D matrix of (N, 6, 6) values are null this will return False and not execute the propagated covariance.

moeyensj: So theoretically, we could check along the first axis and then do some masking to pass only those covariances that are fully defined.

spenczar: > To clarify, if any cells in this 3D matrix of (N, 6, 6) values are null this will return False and not execute the propagated covariance.

Are you sure? I think that's backwards - this is "not all nan", but it sounds like you want "all not nan." (`np.all(~np.isnan(cartesian_covariances))`)

```
>>> import numpy as np
>>> valid = np.ones((6, 6), np.float64())
>>> partial_valid = np.ones((6, 6), np.float64())
>>> partial_valid[0][0] = float('nan')
>>> data = np.array([valid, partial_valid])
>>> print(data)
[[[ 1.  1.  1.  1.  1.  1.]
  [ 1.  1.  1.  1.  1.  1.]
  [ 1.  1.  1.  1.  1.  1.]
  [ 1.  1.  1.  1.  1.  1.]
  [ 1.  1.  1.  1.  1.  1.]
  [ 1.  1.  1.  1.  1.  1.]]

 [[nan  1.  1.  1.  1.  1.]
  [ 1.  1.  1.  1.  1.  1.]
  [ 1.  1.  1.  1.  1.  1.]
  [ 1.  1.  1.  1.  1.  1.]
  [ 1.  1.  1.  1.  1.  1.]
  [ 1.  1.  1.  1.  1.  1.]]]
>>> not np.all(np.isnan(data))
True

moeyensj: Oh s***. You are completely correct. It's because this used to be a masked array and masks are True when they are masked so it's inverted here.. I'll fix it in the 2-body ephemeris generation PR.

moeyensj: Changed here: https://github.com/B612-Asteroid-Institute/adam_core/commit/a580b19ba79f7a9fb27f5cc6e66f96eeff721350 

2-body ephemeris PR isn't ready yet but at least the fix is there.

moeyensj: Actually, `not np.all(np.isnan(cov))` is what I was trying to do. I'm trying to check if any values are not NaN. If any values are not NaN, then lets proceed and do the covariance transformation. If all values are NaN, then the covariance transformation will not yield anything useful so lets skip it.

moeyensj: Proposed solution: #30

PR #24, opened 2023-07-11
Concurrent futures and a few propagation bug fixes

spenczar:
Swapped out multiprocessing for concurrent.futures because it handles crashes better.

Fixed a little latent bug in orbit propagation, and added tests.

Also did a drive-by fixup of Times to let it be constructed from a scalar astropy.time.Time.

### DIFF
--- a/adam_core/propagator/propagator.py
+++ b/adam_core/propagator/propagator.py
@@ -49,7 +49,7 @@ class Propagator(ABC):
         orbits: Orbits,
         times: Time,
         chunk_size: int = 100,
-        num_jobs: Optional[int] = 1,
+        max_processes: Optional[int] = 1,
     ) -> Orbits:
         """
         Propagate each orbit in orbits to each time in times.
@@ -62,9 +62,9 @@ class Propagator(ABC):
             Times to which to propagate orbits.
         chunk_size : int, optional
             Number of orbits to send to each job.
-        num_jobs : int or None, optional
-            Number of jobs to launch. If None then the number of
-            jobs will be equal to the number of cores on the machine. If 1
+        max_processes : int or None, optional
+            Maximum number of processes to launch. If None then the number of
+            processes will be equal to the number of cores on the machine. If 1
             then no multiprocessing will be used.
 
         Returns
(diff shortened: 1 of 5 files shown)


adam_core/propagator/propagator.py
```diff
@@ -145,25 +145,14 @@ def generate_ephemeris(
...
-                    repeat(self),
-                ),
-            )
-            p.close()
-            p.join()
-
+            with concurrent.futures.ProcessPoolExecutor(max_workers=num_jobs) as executor:
+                futures = []
+                for orbit_chunk in _iterate_chunks(orbits, chunk_size):
+                    futures.append(executor.submit(ephemeris_worker, orbit_chunk, observers, self))
```
moeyensj, line 141: I like this interface a lot

adam_core/propagator/utils.py
```diff
@@ -45,8 +45,8 @@ def sort_propagated_orbits(propagated_orbits: Orbits) -> Orbits:
     # Build table with orbit_ids, object_ids, and times
     table = pa.table(
         [
-            propagated_orbits.orbit_ids,
-            propagated_orbits.object_ids,
+            propagated_orbits.orbit_id,
+            propagated_orbits.object_id,
```
moeyensj, line 49: Good catch, I missed this in a previous PR.

moeyensj (approved):
Straightforward! I really like how easy it is to read and understand the concurrent futures code. Very nice.

adam_core/propagator/propagator.py
```diff
@@ -73,24 +73,14 @@ def propagate_orbits(
...
-                propagation_worker,
-                zip(
-                    orbits_split,
-                    repeat(times),
-                    repeat(self),
-                ),
-            )
-            p.close()
-            p.join()
+            with concurrent.futures.ProcessPoolExecutor(max_workers=num_jobs) as executor:
```
akoumjian, line 76: This doesn't really have to do with this PR, but this might be a good opportunity to rename `num_jobs` to `workers` or `processes` or `parallelism` or something equivalent, since a job is a unit of work and the parameter is really about how many separate processes to use.

akoumjian (approved):

adam_core/propagator/propagator.py
spenczar: I like this suggestion a lot, I'll do it here.

PR #26, opened 2023-07-12
Add ephemeris generation 

moeyensj:
This PR adds an observer class, ephemeris class and then ephemeris generation with PYOORB. 

Snippet from the README to get a sense of the interface: 
```python
import numpy as np
from astropy.time import Time
from adam_core.orbits.query import query_horizons
from adam_core.propagator import PYOORB
from adam_core.observers import Observers

# Get orbits to propagate
initial_time = Time([60000.0], scale="tdb", format="mjd")
orbits = query_horizons(object_ids, initial_time)

# Make sure PYOORB is ready
propagator = PYOORB()

# Define a set of observers and observation times
times = initial_time + np.arange(0, 100) * u.d
observers = Observers.from_code("I11", times)

# Generate ephemerides! This function supports multiprocessing for large
# propagation jobs.
ephemeris = propagator.generate_ephemeris(
    orbits,
    observers,
    chunk_size=100,
    num_jobs=1
)
```

### DIFF
--- a/adam_core/observers/observers.py
+++ b/adam_core/observers/observers.py
@@ -1,6 +1,10 @@
 import pandas as pd
+from astropy.time import Time
 from mpc_obscodes import mpc_obscodes
 from quivr import Float64Column, StringColumn, Table
+from typing_extensions import Self
+
+from ..coordinates.cartesian import CartesianCoordinates
 
 
 class ObservatoryGeodetics(Table):
@@ -29,3 +33,112 @@ OBSERVATORY_GEODETICS = ObservatoryGeodetics.from_kwargs(
 OBSERVATORY_CODES = {
     x for x in OBSERVATORY_GEODETICS.code.to_numpy(zero_copy_only=False)
 }
+
+
+class Observers(Table):
+    code = StringColumn(nullable=False)
+    coordinates = CartesianCoordinates.as_column()
+
+    @classmethod
+    def from_code(cls, code: str, times: Time) -> Self:
+        """
+        Instantiate an Observers table with a single code and multiple times.
+        Times do not need to be unique. The observer state will be calculated
+        for each time and correctly matched to the input times and replicated for
+        duplicate times.
+
+        To load multiple codes, use `from_code` and then concatenate the tables.
+
+        Parameters
+        ----------
+        code : str
+            MPC observatory code for which to find the states.
+        times : `~astropy.time.core.Time` (N)
+            Epochs for which to find the observatory locations.
+
+        Returns
+        -------
+        observers : `~adam_core.observers.observers.Observers` (N)
+            The observer and its state at each time.
+
+        Examples
+        --------
+        >>> import numpy as np
+        >>> from astropy.time import Time
+        >>> from adam_core.observers import Observers
+        >>> times = Time(np.arange(59000, 59000 + 100), scale="tdb", format="mjd")
+        >>> observers = Observers.from_code("X05", times)
+        """
+        from .state import get_observer_state
+
+        return cls.from_kwargs(
+            code=[code] * len(times),
+            coordinates=get_observer_state(code, times),
+        )
+
+    def to_dataframe(self) -> pd.DataFrame:
+        """
+        Convert the Observers table to a pandas DataFrame.
+
+        Returns
+        -------
+        df : `~pandas.DataFrame`
+            The Observers table as a DataFrame.
+        """
+        df = self.coordinates.to_dataframe(
+            sigmas=False,
+            covariances=False,
+        )
+        df.rename(
+            columns={
+                "jd1_tdb": "obs_jd1_tdb",
+                "jd2_tdb": "obs_jd2_tdb",
+                "x": "obs_x",
+                "y": "obs_y",
+                "z": "obs_z",
+                "vx": "obs_vx",
+                "vy": "obs_vy",
+                "vz": "obs_vz",
+                "origin.code": "obs_origin.code",
+            },
+            inplace=True,
+        )
+        df.insert(0, "obs_code", self.code)
+        return df
+
+    @classmethod
+    def from_dataframe(cls, df: pd.DataFrame) -> Self:
+        """
+        Instantiate an Observers table from a pandas DataFrame.
+
+        Parameters
+        ----------
+        df : `~pandas.DataFrame`
+            The Observers table as a DataFrame.
+
+        Returns
+        -------
+        observers : `~adam_core.observers.observers.Observers`
+            The Observers table.
+        """
+        df_renamed = df.rename(
+            columns={
+                "obs_jd1_tdb": "jd1_tdb",
+                "obs_jd2_tdb": "jd2_tdb",
+                "obs_x": "x",
+                "obs_y": "y",
+                "obs_z": "z",
+                "obs_vx": "vx",
+                "obs_vy": "vy",
+                "obs_vz": "vz",
+                "obs_origin.code": "origin.code",
+            }
+        )
+        coordinates = CartesianCoordinates.from_dataframe(
+            df_renamed,
+            frame="ecliptic",
+        )
+        return cls.from_kwargs(
+            code=df["obs_code"].values,
+            coordinates=coordinates,
+        )
--- a/adam_core/observers/observers.py
+++ b/adam_core/observers/observers.py
@@ -98,7 +98,7 @@ class Observers(Table):
         observers : `~adam_core.observers.observers.Observers`
             The Observers table for this observer.
         """
-        for code in self.code.unique():
+        for code in self.code.unique().sort():
             yield code.as_py(), self.select("code", code)
 
     def to_dataframe(self) -> pd.DataFrame:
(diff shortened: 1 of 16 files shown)


moeyensj:
I'll follow this up with a PR that has a Photometry class and PhysicalProperties class so we can also add predicted magnitudes from PYOORB to the Ephemeris class.

adam_core/orbits/ephemeris.py
```diff
@@ -0,0 +1,109 @@
...
+    orbit_id = StringColumn(nullable=False)
+    object_id = StringColumn()
+    observer = Observers.as_column(nullable=False)
+    rho = Float64Column()
+    ra = Float64Column(nullable=False)
+    dec = Float64Column(nullable=False)
+    vrho = Float64Column()
+    vra = Float64Column()
+    vdec = Float64Column()
+    frame = StringAttribute()
```
spenczar, line 20: I'm a little surprised this doesn't embed SphericalCoordinates directly as a `SphericalCoordinates.as_column()`. Why not? It would be nice to bring covariances in, and that seems like an easy way.

adam_core/propagator/pyoorb.py
```diff
@@ -282,7 +294,117 @@ def _propagate_orbits(self, orbits: Orbits, times: Time) -> Orbits:
...
+                in_orbits=orbits_pyoorb,
+                in_obscode=code_i,
+                in_date_ephems=epochs_pyoorb,
+                in_dynmodel=self.dynamical_model,
+            )
+            if err != 0:
+                warnings.warn(
+                    f"PYOORB has returned error code {err} while generating ephemerides!",
+                    UserWarning,
+                )
```
spenczar, line 343: Why just warn, and not raise? Is `ephemeris` likely to be a valid value so we're able to proceed anyway?

adam_core/propagator/pyoorb.py
```diff
@@ -282,7 +294,117 @@ def _propagate_orbits(self, orbits: Orbits, times: Time) -> Orbits:
...
+
+            # Stack 3D ephemeris into single 2D array
+            ephemeris = np.vstack(ephemeris)
+
+            # PYOORB returns ephemerides for each orbit, so lets reconstruct orbit IDs
+            ids = np.arange(0, len(orbits))
+            orbit_ids_idx = np.array([i for i in ids for j in range(len(observer_i))])
+            orbit_ids = orbits.orbit_id.to_numpy(zero_copy_only=False)[orbit_ids_idx]
+            object_ids = orbits.object_id.to_numpy(zero_copy_only=False)[orbit_ids_idx]
```
spenczar, line 352: I think the intent here is to repeat each identifier `len(observer_i)` times, right?
```suggestion
            orbit_ids = pa.concat_arrays(itertools.repeat(orbits.orbit_id, len(observer_i)))
            object_ids = pa.concat_arrays(itertools.repeat(orbits.object_id, len(observer_i)))
```

adam_core/propagator/utils.py
```diff
@@ -70,3 +71,51 @@ def sort_propagated_orbits(propagated_orbits: Orbits) -> Orbits:
         ),
     )
     return propagated_orbits.take(indices)
+
+
+def sort_ephemeris(ephemeris: Ephemeris) -> Ephemeris:
```
spenczar, line 76: Some day we'll make a nice API for this. Some day.

adam_core/orbits/ephemeris.py
moeyensj: I'd actually prefer it to be `SphericalCoordinates.as_column()` but what I was thinking was that it makes sense to have a bit of symmetry with our future `Observations` table and `Ephemeris` since we directly compare those for residual calculations. For `Observations` we discussed not having SphericalCoordinates included since for example rho, vrho, vra, vdec will be empty for the vast majority of our observations. Although that said, I guess `Residuals.calculate(Observations.to_spherical_coordinates(), Ephemeris.coordinates)` really isn't that cumbersome?

adam_core/propagator/pyoorb.py
moeyensj: Yes, agreed. Better be conservative and let it fail then let something inaccurate pass through.

adam_core/orbits/ephemeris.py
moeyensj: One other thought too is that times would be duplicated between Observers and SphericalCoordinates. Not sure how to handle that. I think the duplication is okay for now but it would be cool to be able to "link" between tables.

adam_core/propagator/propagator.py
```diff
@@ -127,29 +132,25 @@ def generate_ephemeris(
...
-            ephemeris = self._generate_ephemeris(orbits, observers)
+            return ephemeris_list
 
-        ephemeris.sort_values(
-            by=["orbit_ids", "origin", "times"],
-            inplace=True,
-        )
-        return ephemeris
+        else:
+            return [self._generate_ephemeris(orbits, observers)]
```
moeyensj, line 156: @spenczar Note that right now this function will return a list of MultiKeyLinkages in the absence of having a way to concatenate the linkages.

adam_core/orbits/ephemeris.py
```diff
@@ -0,0 +1,71 @@
...
+
+        Returns
+        -------
+        df : `~pandas.DataFrame`
+            The Ephemeris table as a DataFrame.
+        """
+        df = pd.DataFrame()
+        df["orbit_id"] = self.orbit_id
+        df["object_id"] = self.object_id
+        df_coordinates = self.coordinates.to_dataframe()
```
spenczar, line 26: Probably should `assert self.coordinates.frame == "equatorial"`, or some such

adam_core/propagator/propagator.py
spenczar: With #33 merged, I think you could merge the linkages now.

adam_core/propagator/pyoorb.py
moeyensj: These are RuntimeErrors now.

moeyensj:
@spenczar If you run the example from the README you'll get an error thats related to the time crossmatching: 
```python
import numpy as np
from astropy.time import Time
from adam_core.orbits.query import query_horizons
from adam_core.propagator import PYOORB
from adam_core.observers import Observers

# Get orbits to propagate
initial_time = Time([60000.0], scale="tdb", format="mjd")
orbits = query_horizons(object_ids, initial_time)

# Make sure PYOORB is ready
propagator = PYOORB()

# Define a set of observers and observation times
times = initial_time + np.arange(0, 100) * u.d
observers = Observers.from_code("I11", times)

# Generate ephemerides! This function supports multiprocessing for large
# propagation jobs.
ephemeris = propagator.generate_ephemeris(
    orbits,
    observers,
    chunk_size=100,
    max_processes=1
)
```
```python
ValueError: Ephemerides were not correctly linked to observers. Expected 100 unique keys, got 101.
```

spenczar:
I think I see what's happening. Openorb's fortran code uses 15 decimal-digit

spenczar:
oops

spenczar:
I think I see what's happening. OpenOrb's fortran code uses real numbers with 15 decimal digits of precision[^1]. That's equivalent to a typical 64-bit IEEE754 double. But our `Times` class uses _two_ doubles[^2]. So we call `Times.to_astropy().tt.mjd`[^3] (not all in one place, but that's the idea) in order to convert the two doubles down to a single double array.

When we re-inflate from the single mjd array that comes back out of PYOORB, it's lost some digits of precision - not quite 64 bits of precision, since the SOFA-style pair of doubles isn't a genuine quad-precision float, but some.

I think the right solution is to perform the multikey linkage on a computed 64-bit value instead of the pair of doubles.

---
[^1]: The `Time` fortran class uses `REAL(bp)`: https://github.com/oorb/oorb/blob/master/classes/Time_class.f90#L79-L82 And `bp` is a specification for "base precision", defined here: https://github.com/oorb/oorb/blob/master/classes/Base_class.f90#L58
[^2]: https://github.com/B612-Asteroid-Institute/adam_core/blob/main/adam_core/coordinates/times.py#L11-L12
[^3]: https://github.com/B612-Asteroid-Institute/adam_core/blob/jm/ephemeris/adam_core/propagator/pyoorb.py#L225

adam_core/observers/observers.py
```diff
@@ -29,3 +44,126 @@ class ObservatoryGeodetics(Table):
...
+                "obs_z": "z",
+                "obs_vx": "vx",
+                "obs_vy": "vy",
+                "obs_vz": "vz",
+                "obs_origin.code": "origin.code",
+            }
+        )
+        coordinates = CartesianCoordinates.from_dataframe(
+            df_renamed,
+            frame="ecliptic",
```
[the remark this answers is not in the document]
moeyensj: Yeah, this function definitely isn't careful. I could encode the frame inside the column names for the coordinate dimensions (I have seen others do this before). For cartesian you would have (x_eq, y_eq, z_eq) or (x_ec, y_ec, z_ec). It's hideous but at least it prevents errors. I propose I do this in a separate PR so we keep this one a bit smaller.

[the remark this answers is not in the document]
moeyensj: This has been implemented in #40.

moeyensj:
@spenczar I think this is ready to be merged (and/or reviewed one last time). :)

spenczar (approved):

PR #27, opened 2023-07-25
Upgrade to quivr v0.5.0

spenczar:
Quivr v0.5.0 cleans up some of the FixedSizeList vs List columns constructor signatures.

### DIFF
--- a/adam_core/coordinates/covariances.py
+++ b/adam_core/coordinates/covariances.py
@@ -6,7 +6,7 @@ import pandas as pd
 import pyarrow as pa
 from astropy import units as u
 from astropy.table import Table as AstropyTable
-from quivr import ListColumn, Table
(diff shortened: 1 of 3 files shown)


moeyensj:
Weird how linting errors will cause a failure in CI for PRs but if you look at the newest commit on main there is a green checkmark but linting failed there too.

moeyensj (approved):
Anyway! This looks great, feel free to merge it as is. I can address the linting issue in the ephemeris generation PR.

moeyensj:
> Weird how linting errors will cause a failure in CI for PRs but if you look at the newest commit on main there is a green checkmark but linting failed there too.

I suspect its because the docker caching failed and that doesn't cause the full CI run to be marked as a failure.

PR #30, opened 2023-08-03
Add CoordinateCovariances.is_all_nan()

moeyensj:
To avoid future mistakes when checking if all elements in a stack of covariance matrices are nan, this PR adds a convenience function to `CoordinateCovariances` that will return a boolean.

### DIFF
--- a/adam_core/coordinates/covariances.py
+++ b/adam_core/coordinates/covariances.py
@@ -78,12 +78,35 @@ class CoordinateCovariances(Table):
             Covariance matrices for N coordinates in 6 dimensions.
         """
         # return self.values.combine_chunks().to_numpy_ndarray()
-        cov = np.stack(self.values.to_numpy(zero_copy_only=False))
-        if np.all(cov == None):  # noqa: E711
+        values = self.values.to_numpy(zero_copy_only=False)
+
+        # If all covariance matrices are None, then return a covariances
+        # filled with NaNs.
+        if np.all(values == None):  # noqa: E711
             return np.full((len(self), 6, 6), np.nan)
+
         else:
-            cov = np.stack(cov).reshape(-1, 6, 6)
-        return cov
+            # Try to stack the values into a 3D array. If this works, then
+            # all covariance matrices are the same size and we can return
+            # the stacked matrices.
+            try:
+                cov = np.stack(values).reshape(-1, 6, 6)
+
+            # If not then some of the arrays might be None. Lets loop through
+            # the values and fill in the arrays that are missing (None) with NaNs.
+            except ValueError as e:
+                # If we don't get the error we expect, then raise it.
+                if str(e) != "all input arrays must have the same shape":
+                    raise e
+                else:
+                    for i in range(len(values)):
+                        if values[i] is None:
+                            values[i] = np.full(36, np.nan)
+
+                # Try stacking again
+                cov = np.stack(values).reshape(-1, 6, 6)
+
+            return cov
 
     @classmethod
     def from_matrix(cls, covariances: np.ndarray) -> "CoordinateCovariances":
--- a/adam_core/coordinates/covariances.py
+++ b/adam_core/coordinates/covariances.py
@@ -177,6 +177,17 @@ class CoordinateCovariances(Table):
 
         return cls.from_matrix(covariances)
 
+    def is_all_nan(self) -> bool:
+        """
+        Check if all covariance matrix values are NaN.
+
+        Returns
+        -------
+        is_all_nan : bool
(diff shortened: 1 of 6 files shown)


adam_core/coordinates/tests/test_covariances.py
```diff
@@ -103,3 +103,14 @@ def test_CoordinateCovariances_to_dataframe():
     np.testing.assert_equal(df["sigma_vx"].values, np.sqrt(np.array([25.0, 36.0])))
     np.testing.assert_equal(df["sigma_vy"].values, np.sqrt(np.array([36.0, 49.0])))
     np.testing.assert_equal(df["sigma_vz"].values, np.sqrt(np.array([49.0, 64.0])))
+
+
+def test_CoordinateCovariances_is_all_nan():
```
spenczar, line 108: Might want to add the following test scenarios:
- one of the 6x6 cells is nan, but not all of them
- the list is itself nan, rather than all of its elements being nan (CoordinateCovariances.from_arrays([None, np.ones((6, 6)).flatten()]), i think, but I am not sure)

spenczar (approved):
Good idea!

adam_core/coordinates/tests/test_covariances.py
moeyensj: Done. And it led to finding a missing corner case in `to_matrix()`. See: [97fb521](https://github.com/B612-Asteroid-Institute/adam_core/pull/30/commits/97fb5218e4efa3f0568f3a4fedf0fd82dbce868f).

adam_core/coordinates/covariances.py
```diff
@@ -78,12 +78,35 @@ def to_matrix(self) -> np.ndarray:
             Covariance matrices for N coordinates in 6 dimensions.
         """
         # return self.values.combine_chunks().to_numpy_ndarray()
-        cov = np.stack(self.values.to_numpy(zero_copy_only=False))
-        if np.all(cov == None):  # noqa: E711
+        values = self.values.to_numpy(zero_copy_only=False)
+
+        # If all covariance matrices are None, then return a covariances
+        # filled with NaNs.
+        if np.all(values == None):  # noqa: E711
```
spenczar, line 85: Arrow arrays keep track of the number of null elements they have in a property called `null_count`. That could be quite a lot faster here - it would just look up two integers instead of converting everything to Python objects and comparing item-by-item:

```py
if self.values.null_count == len(self.values):
    return np.full((len(self.values), 6, 6), np.nan)
```

adam_core/coordinates/covariances.py
```diff
@@ -78,12 +78,35 @@ def to_matrix(self) -> np.ndarray:
...
+                    raise e
+                else:
+                    for i in range(len(values)):
+                        if values[i] is None:
+                            values[i] = np.full(36, np.nan)
+
+                # Try stacking again
+                cov = np.stack(values).reshape(-1, 6, 6)
+
+            return cov
```
spenczar, line 109: I think it's better to do most of this work in Arrow space. Here are a few tools you could use.

1. `self.values.value_lengths()` will give you an array of the lengths of each list element. If a list is null, you'll get a null length:
```py
In [11]: a = pa.array([[1, 2, 3], [4, 5, 6], [7, 8, 9], None])

In [12]: a.value_lengths()
Out[12]:
<pyarrow.lib.Int32Array object at 0x1287c7f40>
[
  3,
  3,
  3,
  null
]
```

2. `Array.value_lengths().unique()` does what you'd hope:
```py
In [13]: a.value_lengths().unique()
Out[13]:
<pyarrow.lib.Int32Array object at 0x1287c7e80>
[
  3,
  null
]
```

3. Instead of iterating over the values to find the nulls, you can efficiently get their indexes with `Array.is_null()`:
```
In [42]: a.is_null()
Out[42]:
<pyarrow.lib.BooleanArray object at 0x12f1df2e0>
[
  false,
  false,
  false,
  true
]
```
You can convert that boolean mask to indices with `pa.compute.indices_nonzero`:
```
In [54]: pa.compute.indices_nonzero(a.is_null())
Out[54]:
<pyarrow.lib.UInt64Array object at 0x1381347c0>
[
  3
]
```

I think you could use these to know very efficiently
1. whether reshaping will work
2. whether it would fail because of nulls vs mis-shapen data
3. which values need to be filled

spenczar (commented):
You're too quick for me! I may add these suggestions in a separate PR.

moeyensj:
> You're too quick for me! I may add these suggestions in a separate PR.

Teach me the ways of the arrow, master fletcher!

Happy to do this myself. I can add a benchmark beforehand so we get to see the performance difference.

PR #31, opened 2023-08-04
Make CoordinateCovariances a fixed_shape_tensor

moeyensj:
I was originally going to address @spenczar's excellent comments on #30 and use pyarrow's compute functionality to make `CoordinateCovariances.to_matrix` more efficient. However, it turns out that issue [35599](https://github.com/apache/arrow/issues/35599) has been resolved on the pyarrow side and so now CoordinateCovariances can actually be stored as fixed shape tensors. 

Note that I added benchmarking too. Here are the results before the PR:
Screenshot from 2023-08-04 13-25-35

After the swap to the tensor backed covariance:
Screenshot from 2023-08-04 13-26-00

### DIFF
--- a/adam_core/coordinates/covariances.py
+++ b/adam_core/coordinates/covariances.py
@@ -4,9 +4,9 @@ from typing import Callable, List
 import numpy as np
 import pandas as pd
 import pyarrow as pa
+import quivr as qv
 from astropy import units as u
 from astropy.table import Table as AstropyTable
-from quivr import FixedSizeListColumn, Table
 from scipy.stats import multivariate_normal
 
 from .jacobian import calc_jacobian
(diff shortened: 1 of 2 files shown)


moeyensj:
After some testing with @spenczar:
1. Serialization to and from parquet files is an issue when covariance values are nullable (related to pyarrow issue:https://github.com/apache/arrow/issues/35692). Namely, if a covariance matrix is missing and another one is defined then we get a shape error when trying to load them from a parquet file. 
```python
import pyarrow as pa
import quivr as qv
import numpy as np

class CoordinateCovariances(qv.Table):
   
    values = qv.Column(pa.fixed_shape_tensor(pa.float64(), (6, 6)))

cov = CoordinateCovariances.from_kwargs(values=[None, np.ones((6, 6)).flatten()])
cov.to_parquet('covariances.parquet')

cov2 = CoordinateCovariances.from_parquet('covariances.parquet')
```
```
ArrowInvalid: Expected all lists to be of size=36 but index 1 had size=0
```
2. If we then set it so the values cannot be null this fixes the serialization issue but then we run into https://github.com/apache/arrow/issues/37004 when we allow `CoordinateCovarainces.as_column(nullable=False)`. 
```python
import pyarrow as pa
import quivr as qv
import numpy as np

class CoordinateCovariances(qv.Table):
   
    values = qv.Column(pa.fixed_shape_tensor(pa.float64(), (6, 6)))

class SampleTable(qv.Table):

    covariance = CoordinateCovariances.as_column(nullable=False)

cov = SampleTable.from_kwargs(covariance=CoordinateCovariances.from_kwargs(values=[None, np.ones((6, 6)).flatten()]))
cov.to_parquet('covariances.parquet')

cov2 = CoordinateCovariances.from_parquet('covariances.parquet')
```
```
ArrowInvalid: Column 0: In chunk 0: Invalid: Struct child array #0 invalid: Invalid: Expected 1 child arrays in array of type fixed_size_list[36], got 0
```

Given these two blockers, I'm going to close this PR. And do what I was originally going to do in a separate PR.

moeyensj:
Benchmarking commit was extracted in #32.

PR #33, opened 2023-08-09
Upgrade to quivr v0.6.2

spenczar:
Major changes are:
 - columns are non-nullable by default
 - schemas are checked *much* more rigorously when constructing Table instances (which was the cause of our grief and pain with nullable subtables)
 - columns support default values


moeyensj (approved):

moeyensj:
Nice work!

moeyensj:
The linting issue will be fixed in #28. I'll clean that up and merge it later today. This is good to merge!

PR #35, opened 2023-08-09
Check if times returned from PYOORB are close to the input times

moeyensj:
Requires #33 to be merged to main. Also adds a RuntimeError when PYOORB returns a non-zero error during propagation.


spenczar (approved):
This is ominous! Why would the times be off?

PR #37, opened 2023-08-11
Add PointSourceDetections and Exposures tables

spenczar:
This is a sketch of an idea, more than a solid proposal. I think this would be useful in many, many places.

One question is what the fields of PointSourceDetections should be. Should it hold SphericalCoordinates, actually? Kinda weird, but formally right. It'd have nulls for so many of the values of the table, in practice (rho, vrho, vlon, vlat...). Yeah, radio exists, but it's pretty exceptional for our work.

But maybe there are big advantages to standardizing on SphericalCoordinates here, I'd totally buy that. It might make transforms of the observations much easier.

### DIFF
--- a/adam_core/observations/detections.py
+++ b/adam_core/observations/detections.py
@@ -1,0 +1,79 @@
+from __future__ import annotations
+
+from typing import Iterator
+
+import healpy
+import numpy as np
+import numpy.typing as npt
+import pyarrow
+import pyarrow.compute
+import quivr as qv
+from quivr.validators import and_, ge, le
+
+from ..coordinates import times
+from . import exposures
+
+
+class PointSourceDetections(qv.Table):
+    """
+    PointSourceDetections is a table of data about point source detections.
+
+    """
+
+    id = qv.StringColumn()
+
+    exposure_id = qv.StringColumn(nullable=True)
+
+    # Some, but not all, point source data may include times for
+    # individual observations within an exposure.
+    time = times.Times.as_column(nullable=True)
+
+    # Should this be a SphericalCoordinates instead?
+
+    ra = qv.Float64Column(validator=and_(ge(0), le(360)))
+    ra_sigma = qv.Float64Column(nullable=True)
+
+    dec = qv.Float64Column(validator=and_(ge(-90), le(90)))
+    dec_sigma = qv.Float64Column(nullable=True)
+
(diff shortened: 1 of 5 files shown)


adam_core/observations/exposures.py
```diff
@@ -0,0 +1,74 @@
...
+        unique_codes = self.observatory_code.unique()
+        for code in unique_codes:
+            mask = pc.equal(self.observatory_code, code)
+            yield code, self.apply_mask(mask)
+
+    def observer_states(
+        self,
+        frame: Literal["ecliptic", "equatorial"] = "ecliptic",
+        origin: origin.OriginCodes = origin.OriginCodes.SUN,
+    ) -> cartesian.CartesianCoordinates:
```
spenczar, line 40: I was *really* feeling the power of adam_core when I wrote this method. Nice job!

adam_core/observations/detections.py
```diff
@@ -0,0 +1,79 @@
...
+        Returns an iterator of PointSourceDetections, each grouped by healpixel.
+        """
+        # Gather unique healpixels
+        healpixels = self.healpixels(nside, nest)
+        unique_healpixels = np.unique(healpixels)
+        for healpixel in unique_healpixels:
+            mask = pyarrow.compute.equal(healpixels, healpixel)
+            yield (healpixel, self.apply_mask(mask))
+
+    def link_to_exposures(
```
moeyensj, line 73: This is clean and is something I'll likely imitate for linkage members and similar constructs.

moeyensj (approved):
I'm happy with this. I think we should merge this and play around with it in downstream codes and adjust as necessary.

adam_core/observations/detections.py
```diff
@@ -0,0 +1,79 @@
...
+
+    id = qv.StringColumn()
+
+    exposure_id = qv.StringColumn(nullable=True)
+
+    # Some, but not all, point source data may include times for
+    # individual observations within an exposure.
+    time = times.Times.as_column(nullable=True)
+
+    # Should this be a SphericalCoordinates instead?
```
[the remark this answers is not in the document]
spenczar: I think a `to_spherical_coordinates()` method sounds clearly good. I think we'll know what we want pretty quickly once we get going on range and shift.

PR #38, opened 2023-08-13
Add sigma point sampling and covariance matrix calculation

moeyensj:
This adds a function to sample a covariance matrix with sigma points and provides a function to then reconstruct the mean and covariance matrix from those sigma points. Both are unit tested with real orbits.

### DIFF
--- a/adam_core/coordinates/covariances.py
+++ b/adam_core/coordinates/covariances.py
@@ -14,19 +14,6 @@ from .jacobian import calc_jacobian
 
 logger = logging.getLogger(__name__)
 
-__all__ = [
-    "CoordinateCovariances",
-    "sigmas_to_covariances",
-    "sample_covariance",
-    "transform_covariances_sampling",
-    "transform_covariances_jacobian",
-    "sigmas_to_df",
-    "sigmas_from_df",
-    "covariances_to_df",
-    "covariances_from_df",
-    "covariances_to_table",
-]
-
 COVARIANCE_FILL_VALUE = np.NaN
 
 
@@ -216,12 +203,17 @@ class CoordinateCovariances(Table):
         return np.all(np.isnan(self.to_matrix()))
 
 
-def sample_covariance(
+def sample_covariance_random(
     mean: np.ndarray, cov: np.ndarray, num_samples: int = 100000
-) -> np.ndarray:
+) -> Tuple[np.ndarray, np.ndarray, np.ndarray]:
     """
     Sample a multivariate Gaussian distribution with given
     mean and covariances.
+
+    The returned weights will be equal to 1 / num_samples so that
+    each sample is equally weighted. Weights are returned from this function
+    so its interface is identical to that of
+    `~adam_core.coordinates.covariances.sample_covariance_sigma_points`.
 
     Parameters
     ----------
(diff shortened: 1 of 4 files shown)


adam_core/coordinates/covariances.py
```diff
@@ -212,13 +200,18 @@ def is_all_nan(self) -> bool:
         return np.all(np.isnan(self.to_matrix()))
 
 
-def sample_covariance(
+def sample_covariance_random(
     mean: np.ndarray, cov: np.ndarray, num_samples: int = 100000
 ) -> np.ndarray:
```
spenczar, line 205: Please update this type hint of the return value type

adam_core/coordinates/covariances.py
```diff
@@ -232,10 +225,124 @@ def sample_covariance(
...
+    with given mean and covariances.
+
+    Parameters
+    ----------
+    mean : `~numpy.ndarray` (D)
+        Multivariate mean of the Gaussian distribution.
+    cov : `~numpy.ndarray` (D, D)
+        Multivariate variance-covariance matrix of the Gaussian distribution.
+    alpha : float, optional
+        Spread of the sigma points between 1e^-2 and 1.
```
spenczar, line 254: Wann & Van Der Merwe 2000 suggests 1E-3, is that too low for us for some reason?

adam_core/coordinates/covariances.py
```diff
@@ -232,10 +225,124 @@ def sample_covariance(
...
+
+    # Beta is used to encode prior knowledge about the distribution.
+    # If the distribution is a well-constrained Gaussian, beta = 2 is optimal
+    # but lets set beta to 0 for now which has the effect of not weighting the mean state
+    # with 0 for the covariance matrix. This is generally better for more distributions.
+    beta = 0
+    # Calculate the weights for mean and the covariance matrix
+    # Weight are used to reconstruct the mean and covariance matrix from the sigma points
+    W[0] = lambd / (D + lambd)
+    W_cov[0] = lambd / (D + lambd) + (1 - alpha**2 + beta)
```
spenczar, line 297:
```suggestion
    W_cov[0] = W[0] + (1 - alpha**2 + beta)
```

adam_core/coordinates/covariances.py
```diff
@@ -232,10 +225,124 @@ def sample_covariance(
...
+    D = mean.shape[0]
+
+    # See equation 15 in Wan & Van Der Merwe (2000) [1]
+    N = 2 * D + 1
+    sigma_points = np.empty((N, D))
+    W = np.empty(N)
+    W_cov = np.empty(N)
+
+    # Calculate the scaling parameter lambda
+    lambd = alpha**2 * (D + kappa) - D
```
spenczar, line 284: This is going to be a negative value. 

[Julier & Uhlmann 1997](https://www.cs.unc.edu/~welch/kalman/media/pdf/Julier1997_SPIE_KF.pdf) warns:
image

(they call your lambda "κ" and don't decompose it into alpha and kappa, but I think its the same thing).

I can't tell whether this is an issue, but it seems worth mentioning.

spenczar (approved):
Cautiously approved? This looks like it matches the paper you've cited, although I can't pretend to deeply understand.

This looks extremely numba-able.

adam_core/coordinates/covariances.py
moeyensj: Discussion about alpha values is in #41.

PR #40, opened 2023-08-15
Append coordinate reference frame to dataframe column names 

moeyensj:
This requires #26 to be merged to main.


spenczar (commented):
The tests look good, and the code looks correct, but... yeesh, this is unpleasant stuff.

I suppose this works, but we must keep in mind that this won't get called if the coordinates are used as a subtable, like in `Observers` or whatever.

Perhaps `quivr.to_dataframe` and `quivr.from_dataframe` should support serializing attributes as a (repeated) column in dataframes, including for sub-tables, as part of the flattening process. Something like `coords.to_dataframe(include_attrs=True)`. 

But before doing that, I guess I should ask: what are the scenarios where you are using pandas dataframes? That might help inform the designs.

moeyensj:
> But before doing that, I guess I should ask: what are the scenarios where you are using pandas dataframes? That might help inform the designs.

Mainly for a more eye-friendly inspection of the underlying tables. 

I'm going to close this because it's a bit.. grotesque to encode attributes as column names. I think we should just be diligent to not serialize to/from dataframes in a production setting and instead save things with quivr built-in functionality.

PR #41, opened 2023-08-15
Add VariantOrbits

moeyensj:
VariantOrbits are orbits sampled from an orbit's covariance matrix either via monte-carlo (random) or sigma-point sampling. The idea is that these orbits can then be propagated with a propagator to a desired set of times, and at each time for each unique input orbit the covariance matrix can be reconstruct by "collapsing" the orbits down. 

Here is an example without propagation (as that support will be added in a future PR):
```python
from adam_core.orbits.query import query_sbdb
from adam_core.orbits.variants import VariantOrbits

orbits = query_sbdb(["Duende", "Eros"]) # Length 2 orbits class with covariances
variants = VariantOrbits.create(orbits, method="auto") # Length 26 orbits class with variant orbits (13 sigma points per input orbit)
collapsed_orbits = variants.collapse(orbits) # Length 2 orbits with reconstructed covariance matrix
```

This PR needs #38 to be merged.

### DIFF
--- a/adam_core/coordinates/variants.py
+++ b/adam_core/coordinates/variants.py
@@ -1,6 +1,7 @@
-from typing import Literal, Tuple, Union
+from typing import Generic, Literal, Protocol, TypeVar, Union
 
 import numpy as np
+import pyarrow as pa
 import quivr as qv
 
 from .cartesian import CartesianCoordinates
@@ -22,6 +23,27 @@ CoordinateType = Union[
 ]
 
 
+T = TypeVar("T", bound=CoordinateType, covariant=True)
+
+
+class VariantCoordinatesTable(Generic[T], Protocol):
+    @property
+    def index(self) -> pa.Int64Array:
+        ...
+
+    @property
+    def sample(self) -> T:
+        ...
+
+    @property
+    def weight(self) -> pa.lib.DoubleArray:
+        ...
+
+    @property
+    def weight_cov(self) -> pa.lib.DoubleArray:
+        ...
+
+
 def create_coordinate_variants(
     coordinates: CoordinateType,
     method: Literal["auto", "sigma-point", "monte-carlo"],
@@ -29,7 +51,7 @@ def create_coordinate_variants(
     alpha: float = 1,
     beta: float = 0,
     kappa: float = 0,
-) -> Tuple[np.ndarray, np.ndarray, np.ndarray, CoordinateType]:
+) -> VariantCoordinatesTable[T]:
     """
     Sample and create variants for the given coordinates by sampling the covariance matrices.
     There are three supported methods:
@@ -69,17 +91,8 @@ def create_coordinate_variants(
 
     Returns
     -------
-    idx : '~numpy.ndarray'
-        The index of the coordinate that each sample belongs to.
-    weights : '~numpy.ndarray'
-        Weights of each sample.
-    cov_weights : '~numpy.ndarray'
-        Weights of the samples to reconstruct covariance matrix.
-    samples : {'~adam_core.coordinates.cartesian.CartesianCoordinates',
-                     '~adam_core.coordinates.keplerian.KeplerianCoordinates',
-                     '~adam_core.coordinates.cometary.CometaryCoordinates',
-                     '~adam_core.coordinates.spherical.SphericalCoordinates'}
-        The samples drawn from the coordinate covariance matrices.
+    variants : '~adam_core.coordinates.variants.VariantCoordinates'
+        The variant coordinates.
 
     Raises
     ------
@@ -87,18 +100,29 @@ def create_coordinate_variants(
         If the covariance matrices are all undefined.
         If the input coordinates are not supported.
     """
-    idx_list = []
-    samples_list = []
-    weights_list = []
-    cov_weights_list = []
-    origins_list = []
-    times_list = []
-
     if coordinates.covariance.is_all_nan():
         raise ValueError(
             "Cannot sample coordinate covariances when covariances are all undefined."
         )
 
+    class VariantCoordinates(qv.Table):
+        index = qv.Int64Column()
+        sample = coordinates.as_column()
+        weight = qv.Float64Column()
+        weight_cov = qv.Float64Column()
+
+    if isinstance(coordinates, CartesianCoordinates):
+        dimensions = ["x", "y", "z", "vx", "vy", "vz"]
+    elif isinstance(coordinates, SphericalCoordinates):
+        dimensions = ["rho", "lon", "lat", "vrho", "vlon", "vlat"]
+    elif isinstance(coordinates, KeplerianCoordinates):
+        dimensions = ["a", "e", "i", "raan", "ap", "M"]
+    elif isinstance(coordinates, CometaryCoordinates):
+        dimensions = ["q", "e", "i", "raan", "ap", "tp"]
+    else:
+        raise ValueError(f"Unsupported coordinate type: {type(coordinates)}")
+
+    variants_list = []
     for i, coordinate_i in enumerate(coordinates):
 
         mean = coordinate_i.values[0]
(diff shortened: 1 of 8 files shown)


adam_core/coordinates/variants.py
```diff
@@ -0,0 +1,215 @@
...
+)
+from .keplerian import KeplerianCoordinates
+from .spherical import SphericalCoordinates
+
+CoordinateType = Union[
+    CartesianCoordinates,
+    KeplerianCoordinates,
+    CometaryCoordinates,
+    SphericalCoordinates,
+]
```
spenczar, line 21: Is it really sensible to do this on Cometary and Keplerian elements? That seems like it could be super non-gaussian. Like what does it mean to symmetrically sample from the eccentricity? Just seems risky. I'm a fan of only working with cartesian coordinates internally.

spenczar (commented):
Partial review, but some early comments

adam_core/coordinates/variants.py
akoumjian: This was my only question as well. Does one get less accurate results by forcing the user to convert to Cartesian first (less true to the original coordinates)?

akoumjian (commented):
Leaving a note that I read this over to the best of my limited time and ability. From a software side, consider this approved.

adam_core/orbits/variants.py
```diff
@@ -0,0 +1,177 @@
...
+                    y=orbit.coordinates.y,
+                    z=orbit.coordinates.z,
+                    vx=orbit.coordinates.vx,
+                    vy=orbit.coordinates.vy,
+                    vz=orbit.coordinates.vz,
+                    covariance=CoordinateCovariances.from_matrix(covariance),
+                    frame=orbit.coordinates.frame,
+                    origin=orbit.coordinates.origin,
+                ),
+            )
```
spenczar, line 151: `mean_state` is unused. I think this could be clearer by not assigning to a name ( like `_,  covariance = mean_and_covariance_from_weighted_samples(...)`). 

You could go further, though, and change `mean_and_covariance_from_weighted_samples`. Instead of returning two values (one of which you won't use), you could just have a function called `weighted_covariance` which computes the covariance from samples using weights. You could also have a `weighted_mean` function, of course. 

Those could be improved again by perhaps making them methods on `VariantOrbits`. Then this code might look like:

```py
for key, orbit, variants in link.iterate():
    key = key.as_py()
    covariance = variants.weighted_covariance()
    
```

which feels pretty nice.

In addition, quivr Tables are "copy-on-write". So you could simplify your big copy to make `orbit_collapsed`:

```py
coords = orbit.coordinates
coords.covariance = CoordinateCovariances.from_matrix(covariance)
orbit.coordinates = coordinates
orbits_list.append(orbit)
```

Putting it all together (and dropping the unused `key`):
```suggestion
        for _, orbit, variants in link.iterate():
            assert len(orbit) == 1

            # Use the weighted sample covariance, and leave all other attributes
            # on the orbit unchanged.
            coords = orbit.coordinates
            coords.covariance = CoordinateCovariances.from_matrix(variants.weighted_covariance())
            orbit.coordinates = coordinates
            orbits_list.append(orbit)
```

adam_core/orbits/variants.py
```diff
@@ -0,0 +1,177 @@
...
+        orbits_collapsed = qv.concatenate(orbits_list)
+
+        # Array of indices into the collapsed orbits
+        orbits_idx = pa.array(np.arange(0, len(orbits_collapsed)))
+
+        # Make a list of arrays that will be used to sort the orbits
+        orbits_idx_sorted_list = []
+
+        # Loop over input orbits and figure out where in the collapsed orbits they occur
+        # There has to be an easier or better way to do this?
```
spenczar, line 164: > There has to be an easier or better way to do this?

Yes, there is! Instead of using `link.iterate()`, construct your own keys and query in the order you choose. Something like this:

```py
for orbit in orbits:
    key = link.key(
        orbit_id=orbits.orbit_id[0],
        jd1=orbits.coordinates.time.jd1[0],
        jd2=orbits.coordinates.time.jd2[0],
    )
    variants = link.select_right(key)
    # As above, now:
    coords = orbit.coordinates
    coords.covariance = CoordinateCovariances.from_matrix(variants.weighted_covariance())
    orbit.coordinates = coordinates
    orbits_list.append(orbit)
return qv.concatenate(orbits_list)    
```

adam_core/coordinates/variants.py
moeyensj: > Is it really sensible to do this on Cometary and Keplerian elements? That seems like it could be super non-gaussian. Like what does it mean to symmetrically sample from the eccentricity? Just seems risky. I'm a fan of only working with cartesian coordinates internally.

This is the beauty of sampling. What we really care about is mapping volumes from one space to another and the way we can do this is using sampling methods -- the whole point of the addition of variants! Here are some plots where I've defined a Keplerian orbit with a covariance matrix:
```python
keplerian_coordinates = KeplerianCoordinates.from_kwargs(
    time=Times.from_astropy(
        Time([59000], scale="tdb", format="mjd")
    ),
    a=[2.0],
    e=[0.8],
    i=[10.0],
    raan=[60.0],
    ap=[40.0],
    M=[30.0],
    covariance=CoordinateCovariances.from_sigmas(np.array([[0.1, 0.01, 1, 0.1, 0.01, 0.01]])),
    origin=Origin.from_kwargs(code=["SUN"]),
    frame="ecliptic"
)
```
We can then do two things. 
1. Take the input Keplerian orbit and covariance, sample the Keplerian covariance and then convert each sample to Cartesian and plot those (grey points). 
2. Take the Keplerian orbit and convert it to Cartesian and then sample the Cartesian covariance for variants (red points). 

What we really care about is to make sure that the underlying shapes plotted are approximately the same. Because it means the underlying covariance transformation preserved the shape of the original distribution. This is really what we care about and want to be able to study. There will of course be edgecases like our code will eventually need to be able to handle creating variants at e=0.9 +- 0.2. It needs to know that when e > 1 then a < 0. It doesn't do that now but that is something we can add later when we need it. The conversion code will assert an error when orbits are incorrectly defined. 

keplerian_vs_cartesian_monte-carlo

Also to showcase the insane power of sigma-points sampling. Look how beautiful this is: 
keplerian_vs_cartesian_sigma-point

spenczar: > What we really care about is to make sure that the underlying shapes plotted are approximately the same. Because it means the underlying covariance transformation preserved the shape of the original distribution. This is really what we care about and want to be able to study. 

Yes! These plots are very encouraging, for sure.

> There will of course be edgecases like our code will eventually need to be able to handle creating variants at e=0.9 +- 0.2. It needs to know that when e > 1 then a < 0. It doesn't do that now but that is something we can add later when we need it. The conversion code will assert an error when orbits are incorrectly defined.

Right, this is the sort of issue that I was worried about. You've described a crucial one, which is that the covariance does not really encode complete information about Keplerian parameterizations. Since sigma points rely upon marginal distributions, I'm concerned that we're going to be in for some trouble; forbidden regions for the full vector is a great example.

Another genre of concern though is that this implementation of sigma points assumes that symmetric samples are equally valid: that `e+σ` and `e-σ` should be equally weighted, and are equally representative of the marginal distribution of e. That is, they assume higher-order moments are all zero. I'm concerned that that could be _very_ untrue for us. I'd expect that the likelihood function for `e` has very non-negligible skewness.

(Actually, we should probably be equally concerned for Cartesian parameterizations. The only real symmetric gaussians lying around are probably the astrometric uncertainties of sky-plane coordinates.)

The issue with higher-order moments isn't necessarily fatal. [Julier 1998](https://ui.adsabs.harvard.edu/abs/1998SPIE.3373..271J/abstract)[^1] (cited by van der Merwe & Wan on this exact subject) looks promising, and suggests that you simply need more sigma points. And [Julier & Uhlmann 2004](https://www.cs.ubc.ca/~murphyk/Papers/Julier_Uhlmann_mar04.pdf) has a lot to say about skewness and kurtosis in section 5, "Mitigating Unmodeled Effects".

This is all me running on fear/uncertainty/doubt, though, and if we set `alpha` to a small enough value, maybe we'll be just fine. The plots are definitely very encouraging!

[^1]: Direct link: http://www.gatsby.ucl.ac.uk/~byron/paper.pdf

moeyensj: So I'm playing around with different alpha values. The issue with alpha < 1 is that we get a negative weight for the mean state which is unsupported in `np.cov` so I'll update the code to not use that and instead do the linalg manually.

The second thing is that the unit tests of covariance reconstruction actually aren't testing the real-life use case. We haven't applied a non-linear transformation and are just testing the reconstruction of the input covariance which is probably biasing us to the wrong set of parameters. Setting alpha = 1e-2, beta = 2, kappa = 0 doesn't reproduce the input covariance as well as alpha = 1, beta = 0, kappa = 0, but it still does a decent job (differences at the 1e-15 vs 1e-25 level, basically 0 for both). You are correct that we will likely want the former as it captures more moments of the distribution (since beta is a function of the kurtosis in the case of Gaussians). 

I'll add a pathway for the additional mitigations from [Julier & Uhlmann 2004](https://www.cs.ubc.ca/~murphyk/Papers/Julier_Uhlmann_mar04.pdf) since they are fairly straightforward for kurtosis. I need to let the statements about skewness marinate a little more -- it's not obvious we can do much there or at least find a one-size-fits-all solution. 

I'm going to document this a bit for the THOR use case since we will want to investigate this further which we can do with simulation. What we have could be good enough as it is for observation gathering but it's better to be sure.

moeyensj: I'm going to continue this part of the work in a separate PR as this one is large enough.

adam_core/orbits/variants.py
moeyensj: Sorted iteration like this is super nice! Implemented.

PR #44, opened 2023-08-19
Support transforming to cartesian coordinates

spenczar:
Needs a test.

PR #49, opened 2023-08-23
Add unit tests for rotations between frames

moeyensj:
..and fixes a pretty serious bug in the rotation function.


moeyensj:
This PR requires #44 to be merged. I'll add a unit test to this PR since the test data added here will be useful to test #44.

PR #53, opened 2023-08-28
Add the aberrated orbit to the Ephemeris table

moeyensj:
Getting the light-time corrected orbit as part of the Ephemeris table is necessary for THOR and likely quite helpful for other use cases.

### DIFF
--- a/adam_core/orbits/ephemeris.py
+++ b/adam_core/orbits/ephemeris.py
@@ -1,6 +1,4 @@
-import pandas as pd
 from quivr import StringColumn, Table
-from typing_extensions import Self
 
 from ..coordinates.spherical import SphericalCoordinates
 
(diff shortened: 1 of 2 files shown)


moeyensj:
I also removed the dataframe serialization for the Ephemeris table since it's unused and best to avoid.

adam_core/orbits/ephemeris.py
```diff
@@ -1,71 +1,12 @@
...
-            ),
-            "equatorial",
-        )
-        return cls.from_kwargs(
-            orbit_id=df["orbit_id"],
-            object_id=df["object_id"],
-            coordinates=coordinates,
-        )
+    coordinates = SphericalCoordinates.as_column()
+    aberrated_orbit = CartesianCoordinates.as_column()
```
moeyensj, line 12: Happy to also call it aberrated_coordinates instead of orbit to avoid confusion with the Orbits table.

moeyensj: Renamed, comment added, and made nullable!

moeyensj:
Yeah, it is a bit quirky to require all propagators to return this. We could calculate the state from the spherical ephemeris if it's missing. One thing I will do for the 2-body ephemeris code is return the aberrated coordinates and the light travel time so we will have that there as well. FindOrb allows you to ask for either Cartesian or Spherical ephemeris so that's an option maybe. We will need to see as we expand propagator support.

PR #55, opened 2023-08-29
Unit tests for CartesianCoordinate attributes, translation, and rotations

moeyensj:
This PR adds some much-needed unit tests for the Cartesian coordinate `.translate` and `.rotate` functions (these are already tested in other places but not with simple translations and matrices). Critically, it also fixes a bug in the `.sigma_r` and `.sigma_v` attributes.


spenczar (approved):
I love it when tests catch a bug!

PR #56, opened 2023-08-30
Add ephemeris test data

moeyensj:
This PR adds ephemeris test data for each of our test objects in a window +- 30 days from the epoch at which the state vectors were gathered.


moeyensj:
I might want to expand this to include 2-3 observations per night. Not just 1 per night.

adam_core/utils/helpers/data/get_test_data.py
```diff
@@ -141,3 +147,50 @@ def _get_orbital_elements(
...
+        ephemeris["orbit_id"] = [f"{i:05d}" for _ in range(len(times))]
+
+        propagated_orbits_list.append(propagated_orbit)
+        ephemeris_dfs.append(ephemeris)
+
+    propagated_orbits: Orbits = qv.concatenate(propagated_orbits_list)
+    propagated_orbits.to_dataframe().to_csv(
+        files("adam_core.utils.helpers.data").joinpath("propagated_orbits.csv"),
+        index=False,
+    )
```
spenczar, line 191: Why the intermediate to_dataframe, instead of straight to CSV?

spenczar (approved):

adam_core/utils/helpers/data/get_test_data.py
moeyensj: The main reason is I didn't know `qv.Table` had a `.to_csv`. But also, the adam_core orbits class has its own definition of `.to_dataframe` that flattens the covariance matrix into a lower triangular representation with appropriate column names. Not sure how much we care about the legibility of the testing files though.

moeyensj: It looks like I pulled some shenanigans in the Times class as well that [encodes the time scale as part of the column names](https://github.com/B612-Asteroid-Institute/adam_core/blob/main/adam_core/coordinates/times.py#L61). I think we might want to revisit some of the decisions I've made with regard to `.to_dataframe` and `.from_dataframe` in some of the adam_core classes. I think it's likely that `to_csv` using quivr is easy and legible enough that we could go through and remove the dataframe implementations here.

PR #57, opened 2023-08-30
Use signed integers for dictionary keys in Exposures

spenczar:
Necessary because of https://github.com/apache/arrow/issues/37476 unfortunately.


moeyensj (approved):
Well caught! 

In the future when we work on photometry fitting, we may want to explicitly define our own mapping from filter name to integer in the form of an enum or something (to lay out which filters we currently support for filter conversion). This could influence how we define this dictionary column. Obviously, that is waaay out of scope for this bug fix, but I note it here.

PR #60, opened 2023-09-12
Add unit tests for imports of key adam_core functions

moeyensj:
I'm finding it fairly tedious to do module-level imports to avoid circular import errors. I'm hitting a lot of overlap with variable names in the existing code and the names of the modules. This isn't an issue on main at the moment but I have run into it a few times in feature branches. 

Example:
```python
from adam_core import observers

# Here we overwrite the module with the object
observers = observers.Observers.from_kwargs(...)
```

The cleanliness of the following seems a lot nicer:
```python
from adam_core.observers import Observers

observers = Observers.from_kwargs(...)
```

This PR adds a unit test to make sure that the main functions in adam_core can be imported as objects.

### DIFF
--- a/adam_core/dynamics/__init__.py
+++ b/adam_core/dynamics/__init__.py
@@ -3,5 +3,6 @@ from .barker import solve_barker
 from .chi import calc_chi
 from .kepler import solve_kepler
 from .lagrange import apply_lagrange_coefficients, calc_lagrange_coefficients
+from .propagation import propagate_2body
 from .stumpff import calc_stumpff
 from .tisserand import calc_tisserand_parameter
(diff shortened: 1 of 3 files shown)


moeyensj:
Obviously, this will mean we will need to do some additional due diligence to make sure that future additions to the code avoid circular imports. 

@spenczar and @akoumjian, your thoughts on this would be extremely welcome.

spenczar:
Yeah, I have been accidentally shadowing module names too.

The tests seem like an OK way to detect circular imports. I worry that they won't help resolve them. But I am ok with going back to importing objects.

moeyensj:
> The tests seem like an OK way to detect circular imports. I worry that they won't help resolve them. But I am ok with going back to importing objects.

100% agree.

akoumjian:
Yes, or changing the module names on import `import adam_core.observers as observers_mod`. Not pretty. Object imports work for me.

PR #62, opened 2023-09-15
Add Associations table

moeyensj:
This PR also adds an Associations table that gets returned with `~adam_core.utils.helpers.observations.make_observations`. This isn't ready to be merged yet.

### DIFF
--- a/adam_core/observations/associations.py
+++ b/adam_core/observations/associations.py
@@ -20,15 +20,28 @@ class Associations(qv.Table):
 
     def group_by_object(self) -> Iterator["Associations"]:
         """
-        Returns an iterator of PointSourceDetections, each grouped by exposure_id.
+        Group the associations by object ID. Any null object IDs will be returned last.
+
+        Returns
+        -------
+        associations : Iterator[`~adam_core.observations.associations.Associations`]
+            Associations grouped by object ID.
         """
         # Gather unique exposure IDs
         object_ids = self.object_id.unique()
         sorted = self.table.sort_by("object_id")
-        for object_id in object_ids:
+
+        # Return non-null object IDs first
+        for object_id in pc.drop_null(object_ids):
             mask = pc.equal(sorted.column("object_id"), object_id)
             table = sorted.filter(mask)
-            yield Associations(table)
+            yield Associations.from_pyarrow(table)
+
+        # If there are any null object IDs, return them last
+        if object_ids.null_count > 0:
+            mask = pc.is_null(sorted.column("object_id"))
+            table = sorted.filter(mask)
+            yield Associations.from_pyarrow(table)
 
     def link_to_detections(
         self, detections: PointSourceDetections
--- a/adam_core/observations/associations.py
+++ b/adam_core/observations/associations.py
@@ -1,0 +1,45 @@
+from __future__ import annotations
+
+from typing import Iterator
+
+import pyarrow.compute as pc
+import quivr as qv
+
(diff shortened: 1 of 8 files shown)


adam_core/observations/tests/test_associations.py
```diff
@@ -0,0 +1,47 @@
...
+    groups = list(associations.group_by_object())
+    assert len(groups) == 3
+    assert groups[0].object_id.to_pylist() == ["o1", "o1"]
+    assert groups[0].detection_id.to_pylist() == ["d1", "d2"]
+
+    assert groups[1].object_id.to_pylist() == ["o2", "o2"]
+    assert groups[1].detection_id.to_pylist() == ["d3", "d4"]
+
+    # What is the expected return here?
+    assert groups[2].object_id.to_pylist() == [None]
```
moeyensj, line 47: @spenczar For the associations table I think it might be reasonable to expect that sometimes we will have unassociated observations (represented as nulls?). If we wanted to group by object ID, is there a way that quivr can return the null group? Right now this test fails. Note that the length of the groups list is 3.

adam_core/observations/associations.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def group_by_object(self) -> Iterator["Associations"]:
+        """
+        Returns an iterator of PointSourceDetections, each grouped by exposure_id.
+        """
+        # Gather unique exposure IDs
+        object_ids = self.object_id.unique()
+        sorted = self.table.sort_by("object_id")
+        for object_id in object_ids:
+            mask = pc.equal(sorted.column("object_id"), object_id)
```
spenczar, line 29: If `object_id` is `null`, then the equality check doesn't yield True, nor False - it returns `null`. Then your filter will not match anything, since the mask is a big array of nulls.

I think you'll need special handling for a null object_id, therefore.

To avoid checking if `object_id` is null on every iteration (since it almost _never_ is null), you could do something like this:

```py
if object_ids.null_count > 0:
    # There is a null group too
    table = sorted.filter(pc.is_null(sorted))
    yield Associations.from_pyarrow(table)
    ```

File: adam_core/observations/associations.py [hunk 3]
```diff
@@ -0,0 +1,45 @@
...
+            table = sorted.filter(mask)
+            yield Associations(table)
```
spenczar, line 31:
```suggestion
            yield Associations.from_pyarrow(table)
```

spenczar (changes requested):

adam_core/observations/tests/test_associations.py
spenczar: See https://github.com/B612-Asteroid-Institute/adam_core/pull/62#discussion_r1327613037

adam_core/observations/associations.py
moeyensj: Fixed in [e120b9a](https://github.com/B612-Asteroid-Institute/adam_core/pull/62/commits/e120b9a8fb2d80fd86795dea2558ad8db1271280).

adam_core/observations/associations.py
moeyensj: Thanks for the explanation! That makes sense. 

Implemented your suggestion in: [e120b9a](https://github.com/B612-Asteroid-Institute/adam_core/pull/62/commits/e120b9a8fb2d80fd86795dea2558ad8db1271280). Let me know if this is what you were thinking.

spenczar (approved):
LGTM

PR #63, opened 2023-09-15
Allow transform_coordinates to handle MPC observatory codes as input …

moeyensj:
…origins

### DIFF
--- a/adam_core/coordinates/transform.py
+++ b/adam_core/coordinates/transform.py
@@ -1215,13 +1215,23 @@ def cartesian_to_origin(
     coords : `~adam_core.coordinates.cartesian.CartesianCoordinates`
         Cartesian coordinates and optionally their covariances.
     origin : `~adam_core.coordinates.origin.OriginCodes`
-        Name of the desired origin.
+        Desired origin. Input origins may be either `~adam_core.coordinates.origin.OriginCodes`
+        or a str of an observatory code, but the output origin (this kwarg) should always be an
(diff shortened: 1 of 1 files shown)


adam_core/coordinates/transform.py
```diff
@@ -1231,12 +1241,28 @@ def cartesian_to_origin(
...
+
+            vectors[mask] = get_observer_state(
+                origin_in.as_py(),
+                times[mask],
+                frame=coords.frame,
+                origin=origin,
+            ).values
+
+        else:
+            raise ValueError("Unsupported origin: {}".format(origin_in.as_py()))
```
moeyensj, line 1265: How many times can I repeat: `origin_in.as_py()` ... I should really just have done: `origin_str = orgin_in.as_py()` once.

adam_core/coordinates/transform.py
```diff
@@ -1231,12 +1241,29 @@ def cartesian_to_origin(
...
-        vectors[mask] = get_perturber_state(
-            OriginCodes[origin_in.as_py()],
-            times[mask],
-            frame=coords.frame,
-            origin=origin,
-        ).values
+
+        origin_in_str = origin_in.as_py()
+        # Could use try / except block here but this is more explicit
+        if origin_in_str in OriginCodes.__members__:
```
spenczar, line 1247: TIL `__members__` is in the public API of enum, weird. But works fine!

spenczar (approved):

PR #64, opened 2023-09-20
Add sort_by to support common sorts

moeyensj:
While working on THOR v2.0, I'm finding myself doing a lot of sorts, especially involving time. This PR adds a `sort_by` function to a couple of the Tables that I'm sorting the most frequently. Note that when sorting by time we need to create a composite column that adds jd1 and jd2, hence why quivr's sort_by doesn't work out of the box here.

### DIFF
--- a/adam_core/observers/observers.py
+++ b/adam_core/observers/observers.py
@@ -1,7 +1,9 @@
 import warnings
-from typing import Union
+from typing import List, Union
 
 import pandas as pd
+import pyarrow as pa
+import pyarrow.compute as pc
 from astropy.time import Time
 from mpc_obscodes import mpc_obscodes
 from quivr import Float64Column, StringColumn, Table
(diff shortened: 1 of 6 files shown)


moeyensj:
@spenczar Do you think there is a way for quivr to enable sorts for nested tables?

```python
orbits = orbits.sort_by(by=["coordinates.time.jd1", "coordinates.time.jd2"], ascending=True)
```

That would effectively get rid of this PR. 

I guess it would be a list of tuples with sort order included (following what pyarrow does):
```python
orbits = orbits.sort_by([("coordinates.time.jd1", "ascending"), ("coordinates.time.jd2", "ascending")])
```

moeyensj:
I've created a proposed implementation in quivr: https://github.com/spenczar/quivr/pull/54

moeyensj:
This is no longer needed now that `quivr` [0.7.0](https://github.com/spenczar/quivr/compare/v0.6.6...v0.7.0) supports dot-delimited sorting for subtables.

PR #67, opened 2023-09-26
Port over the logic for dataframe conversions to stuff time scales into DFs

spenczar:
This is just a copy of existing code used for coordinates.time.Time tables.

Note that this does _not_ work when the times are used as a subcolumn; see the failing test.

One way we could manage this would, perhaps, be to use DataFrame attrs: https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.attrs.html We could stash the scalar attributes in that dictionary, and pull em back out. That could be a quivr feature.

### DIFF
--- a/adam_core/time/tests/test_time.py
+++ b/adam_core/time/tests/test_time.py
@@ -437,3 +437,38 @@ class TestTimeMath:
             86400_000_000_000 - 199,
             86400_000_000_000 - 298,
         ]
+
+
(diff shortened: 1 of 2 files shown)


spenczar:
> Maybe quivr needs a nice __repr__ for tables?

I'm in favor of this. Perhaps via the direct IPython hooks: https://ipython.readthedocs.io/en/stable/config/integrating.html#rich-display

PR #70, opened 2023-09-27
Write test orbit data as parquet instead of CSV

spenczar:
This is more robust and removes any dependency on custom dataframe and CSV handling.


moeyensj (approved):
These are good changes!

Issue #71, opened 2023-09-27
Orbits with null covariances cannot be loaded directly from parquet files


moeyensj:
To reproduce the issue: 

```python
from adam_core.orbits import Orbits
from adam_core.utils.helpers import make_real_orbits

orbits = make_real_orbits()
orbits = orbits.set_column("coordinates.covariance.values", None)
orbits.to_parquet("orbits.parquet")

orbits = Orbits.from_parquet("test.parquet")
```
Which raises a ArrowInvalidError:

```python
---------------------------------------------------------------------------
ArrowInvalid                              Traceback (most recent call last)
Cell In[6], line 1
----> 1 orbits.from_parquet("test.parquet")

File ~/software/anaconda3/envs/thor_py310/lib/python3.10/site-packages/quivr/tables.py:797, in Table.from_parquet(cls, path, memory_map, pq_buffer_size, filters, column_name_map, validate, **kwargs)
    765 @classmethod
    766 def from_parquet(
    767     cls,
   (...)
    774     **kwargs: AttributeValueType,
    775 ) -> Self:
    776     """Read a table from a Parquet file.
    777 
    778     :param path: The path to the Parquet file.
   (...)
    795 
    796     """
--> 797     table = cls._load_parquet_table(
    798         path=path,
    799         memory_map=memory_map,
    800         pq_buffer_size=pq_buffer_size,
    801         filters=filters,
    802         column_name_map=column_name_map,
    803     )
    804     return cls.from_pyarrow(table=table, validate=validate, permit_nulls=False, **kwargs)

File ~/software/anaconda3/envs/thor_py310/lib/python3.10/site-packages/quivr/tables.py:830, in Table._load_parquet_table(cls, path, memory_map, pq_buffer_size, filters, column_name_map)
    827     column_names = [field.name for field in cls.schema]
    828     schema = cls.schema
--> 830 table = pyarrow.parquet.read_table(
    831     source=path,
    832     columns=column_names,
    833     memory_map=memory_map,
    834     buffer_size=pq_buffer_size,
    835     filters=filters,
    836     schema=schema,
    837 )
    838 md = pyarrow.parquet.read_metadata(path, memory_map=memory_map)
    839 table = table.replace_schema_metadata(md.metadata)

[...]

PR #72, opened 2023-09-27
No more custom dataframes

moeyensj:
This PR removes all the custom to/from_dataframe methods -- feels so much cleaner not to override what quivr is doing. 

These tests will fail until we get a fix in for #71.


moeyensj:
I've modified the Observer states test data to be stored as parquet files as well with null-filled covariances. This allows the unit tests to pass for those functions.

Despite having removed the custom dataframes for Timestamp, I am still getting errors for the dataframe roundtripping.

moeyensj:
The error I'm getting on the last timestamp round trip test is the following: 
```python
kwargs = {'scale': 'tdb'}, name = 'scale', value = 'tdb'

    def __init__(self, table: pa.Table, **kwargs: AttributeValueType):
        self.table = table
        for name, value in kwargs.items():
            if name in self._quivr_attributes:
                setattr(self, name, value)
            else:
>               raise AttributeError(f"{self.__class__.__name__} has no attribute {name}")
E               AttributeError: Wrapper has no attribute scale

../quivr/quivr/tables.py:138: AttributeError
```
Looks like attributes are not passed down to nested tables when using from_flat_dataframe? Is this the correct behavior?

I definitely like the idea of trying out dataframe attributes as a way to persist quivr attributes.

spenczar:
Correct, dataframes do not store attributes, so they won't round-trip.

`to_csv` has special behavior for this, which should perhaps be ported over to the dataframes: https://github.com/spenczar/quivr/blob/main/quivr/tables.py#L875-L880

moeyensj:
Thanks for doing that, I'll re-mark the pytest as an expected failure for the time being.

spenczar:
The dataframes should preserve attributes as of quivr v0.7.0a3. Maybe its time to release 0.7.0 though.

moeyensj:
@spenczar Unit tests pass with the pre-release version (https://github.com/B612-Asteroid-Institute/adam_core/pull/72/commits/7a8a6b775c64cb336a14b6c1f0519b72b60c3755). Thanks for getting dataframe attributes implemented.

I can get [the sort_by PR](https://github.com/spenczar/quivr/pull/54) review comments addressed in the next few hours if you would rather wait to roll that into 0.7.0. No preference on my side.

PR #73, opened 2023-10-02
Use integer time throughout codebase

spenczar:
This is a big PR. I'm relying very heavily on tests, and there are probably spots that I have missed.

I'll try to highlight areas of special concern.


adam_core/coordinates/cometary.py
```diff
@@ -262,7 +262,7 @@ def from_cartesian(cls, cartesian: CartesianCoordinates) -> "CometaryCoordinates
 
         coords_cometary = cartesian_to_cometary(
             cartesian.values,
-            cartesian.time.to_astropy().tdb.mjd,
+            cartesian.time.to_numpy(),
```
spenczar, line 265: I've decided to add to_numpy as a convenience method for the conversion to TDD-scaled, MJD-epoched, double-precisioned timestamps. Perhaps the name is opaque?

adam_core/coordinates/covariances.py
```diff
@@ -153,6 +153,25 @@ def is_all_nan(self) -> bool:
         """
         return np.all(np.isnan(self.to_matrix()))
 
+    @classmethod
+    def nulls(cls, length: int) -> "CoordinateCovariances":
```
spenczar, line 157: This function is helpful when creating a "null" covariances which can be written to and read from Parquet, because https://github.com/apache/arrow/issues/35692 remains an issue.

adam_core/dynamics/propagation.py
```diff
@@ -163,7 +163,7 @@ def propagate_2body(
             vy=orbits_propagated[:, 4],
             vz=orbits_propagated[:, 5],
             covariance=cartesian_covariances,
-            time=Times.from_astropy(
+            time=Timestamp.from_astropy(
                 Time(t1_, scale="tdb", format="mjd"),
             ),
```
spenczar, line 168: There are a _lot_ of `Timestamp.from_astropy(astropy.Time(...))` calls littered throughout the codebase. Those should be shrinkable down to just `Timestamp(...)` in some way - possibly a new direct constructor. For example here it would be `Timestamp.from_mjd(t1_, scale="tdb"). 

I could do those now, if that seems useful.

adam_core/observers/tests/test_state.py
```diff
@@ -37,191 +37,49 @@ def test_get_observer_state_OriginCodes():
...
+    "I41": {"r": 10, "v": 2},
+    "W84": {"r": 10, "v": 2},
+    "000": {"r": 10, "v": 2},
+    "500": {"r": 0.10, "v": 0.01},
+}
+
+
+@pytest.mark.parametrize("code", ["X05", "I41", "W84"])
+@pytest.mark.parametrize("origin", ["sun", "ssb"])
+def test_get_observer_state(code, origin):
```
spenczar, line 51: This file looks like it got a lot of changes, but really, this is just parametrization to reduce the repetition. It doesn't do anything novel.

adam_core/observers/tests/testdata/get_states.py
```diff
@@ -0,0 +1,48 @@
+import numpy as np
```
spenczar, line 1: This file's contents come from the README, more or less.

adam_core/orbits/variants.py
```diff
@@ -104,13 +104,13 @@ def link_to_orbits(
             self,
             left_keys={
                 "orbit_id": orbits.orbit_id,
-                "jd1": orbits.coordinates.time.jd1,
-                "jd2": orbits.coordinates.time.jd2,
+                "day": orbits.coordinates.time.days,
+                "millis": orbits.coordinates.time.millis(),
```
spenczar, line 108: This is a consequential choice - what's the linking precision for these variants? I figured milliseconds is good enough, but maybe micros? Maybe seconds? Maybe nanos? You tell me.

adam_core/coordinates/cometary.py
moeyensj: Another option (now having read more of the code) might be to enable symmetry between `.from_mjd` by adding a `to_mjd(scale="tdb")`. The current `mjd()` could suffice as well with a scale parameter. The latter could just return a pyarrow array (the additional call to `.to_numpy()` from there doesn't seem super cumbersome from my perspective). 

I suspect a lot of astronomers will expect an interface similar to astropy. Don't know how much we want to mimic it but that's a thought. This doesn't have to be addressed in this PR obviously. Here are some options:

```python
import numpy as np
from adam_core.time import Timestamp

mjd_tdb = np.arange(59000, 60000)

times = Timestamp.from_mjd(mjd_tdb, scale="tdb") # this feels very nice and not much different than astropy
mjd_utc = times.utc().mjd().to_numpy() 
mjd_utc = times.mjd(scale="utc").to_numpy() 
mjd_utc = times.to_numpy(scale="utc")
mjd_utc = times.to_mjd(scale="utc") # the symmetry between .from_mjd() and .to_mjd() feels quite intuitive
```

moeyensj (commented):
Early review comments and please push back where necessary. 

One thing to check so we avoid continually reverting/redoing changes to the test data is if your astropy cache is affecting the test data diffs.  

The only other "major" thing is the interface of the Timestamp class. This will likely replace our use of astropy time so it would be good I think to come up with an API we are happy with. I think astronomers will expect to see something astropy-esque. Having that level of convenience and ease of use would be nice. We may want to consult @mjuric, or better maybe we get you to do a quick show and tell in the DiRAC Solar System meeting to get opinions there.  

Lastly, I think we should go "all-in" and change the interfaces of our primary functions (the ones in `test_imports.py` that are not jax-jitted like `propagator.propagate_orbits`) to accept `Timestamp` objects and not astropy Time objects. This could be done in a separate PR if needed but it would inform how much we like the interface to the Timestamp class.

moeyensj:
I forgot to say this in the review: These are excellent changes and I'm really excited to be able to move away from insane floating point time madness.

adam_core/dynamics/propagation.py
akoumjian: Yes, definitely.

adam_core/observations/exposures.py
```diff
@@ -19,7 +18,7 @@ class Exposures(qv.Table):
     """
 
     id = qv.StringColumn()
-    start_time = Times.as_column()
+    start_time = Timestamp.as_column()
     duration = qv.Float64Column(validator=and_(ge(0), le(3600)))
```
akoumjian, line 22: I wonder if switching this to an integer (nanoseconds?) would be worth the inconvenience?

adam_core/time/interval.py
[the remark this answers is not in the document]
spenczar: Oops! Intervals are an idea to support differencing but I don't think they're ready for action.

PR #77, opened 2023-10-10
Vectorize Origin.mu and use throughout adam_core

moeyensj:
This PR cleans up how the gravitational parameter as calculated from an origin code is used. We had inconsistent behavior in adam_core where some functions would use the constants module's solar GM despite the origin conceivably being something else. Instead, the keplerian and cometary conversion functions now require mu to be explicitly passed, and the `propagate_2body` and `generate_ephemeris_2body` now will calculate mu directly from the input orbits.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,10 +68,10 @@ orbits = Orbits.from_kwargs(
 
 Orbits can be easily converted to a pandas DataFrame:
 ```python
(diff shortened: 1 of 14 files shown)


moeyensj:
Excellent point, let me do that now. There are only a couple of examples that are affected by this change.

PR #78, opened 2023-10-10
Add time linkage with precision

moeyensj:
This PR adds a `Timestamp.link` function that lets you link to another `Timestamp` with the user-desired precision.

### DIFF
--- a/adam_core/time/tests/test_time.py
+++ b/adam_core/time/tests/test_time.py
@@ -502,3 +502,58 @@ class TestUnique:
         assert len(have) == 4
         pairs = zip(have.days.to_pylist(), have.nanos.to_pylist())
         assert sorted(list(pairs)) == [(1, 2), (1, 3), (2, 4), (2, 5)]
+
+
+@pytest.mark.parametrize("rescale", [True, False])
+def test_Timestamp_link(rescale):
+    # Test that the link method works as expected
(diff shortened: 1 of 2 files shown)


adam_core/time/tests/test_time.py
```diff
@@ -502,3 +502,108 @@ def test_multple(self):
...
+@pytest.mark.parametrize("rescale", [True, False])
+def test_Timestamp_link(rescale):
+    # Test that the link method works as expected
+    time = Timestamp.from_kwargs(
+        days=[1, 1, 2, 2, 3, 3], nanos=[1, 2, 3, 3, 4, 4], scale="tdb"
+    )
+
+    other = Timestamp.from_kwargs(days=[1, 1, 2, 3], nanos=[1, 2, 3, 4], scale="tdb")
+    if rescale:
+        other = other.rescale("utc")
```
spenczar, line 516: UTC isn't well-defined before about 1960. Could you use some `days` values which would trigger a more substantial rescaling? Stuff around `60000` is all good.

spenczar (approved):

adam_core/time/tests/test_time.py
moeyensj: Great suggestion. Changed these dates to `68000`+.

PR #81, opened 2023-10-13
Return Ephemeris from generate_ephemeris functions

moeyensj:
This PR changes the generate ephemeris function to return the Ephemeris table. A function `Ephemeris.link_to_observers` was added to support linking back to the input observers.


akoumjian (approved):
I like it. Much cleaner.

PR #83, opened 2023-10-16
API documentation generation

spenczar:
Docs visible here: https://adam-core--83.org.readthedocs.build/en/83/

This is the scaffolding for automatic documentation generation from Python docstrings. We can use this to make the API reference for adam_core.

I have expressly **not** attempted to clean up all the existing docstrings, type annotations, etc. I did a few when they caused egregious issues, like inability to build. But I'd like to revise _content_ in separate PRs. 

The objects that get documented are those in the `__all__` list for each module. That is especially relevant to `adam_core.coordinates`. The `__init__` file for that module has a _lot_ of stuff it imports, but most of that stuff looks internal (to me) so I've only chosen to identify a few objects for the public API.


moeyensj (approved):
I don't really have much feedback here. This is excellent! I'm a huge fan of the links created for function parameters when browsing the documentation site.

PR #85, opened 2023-10-18
Delete unused code and remove private stuff from public APIs

spenczar:
As discussed in call


adam_core/coordinates/cartesian.py
```diff
@@ -14,16 +13,13 @@
     from .keplerian import KeplerianCoordinates
     from .spherical import SphericalCoordinates
 
-__all__ = ["CartesianCoordinates", "CARTESIAN_COLS", "CARTESIAN_UNITS"]
+__all__ = ["CartesianCoordinates"]
 
 CARTESIAN_COLS = {}
```
moeyensj, line 18: I suspect that these `{CoordinateType}_COLS` dictionaries are also no longer being used, so they might be safe to remove.

moeyensj (approved):
Always a good day when 700+ lines of code get binned!

PR #88, opened 2023-10-20
Import astropy.time in time.py

moeyensj:
This circumvents an issue with astropy's import of the time subpackage. At import time, astropy.time will check for cached files and download files such as the leapseconds definitions. This only occurs if you explicitly import astropy.time and not astropy.

### DIFF
--- a/adam_core/time/time.py
+++ b/adam_core/time/time.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-import astropy
+import astropy.time
 import numpy as np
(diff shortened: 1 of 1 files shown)


akoumjian:
I think this fixes it only because I am unable to reproduce the bug anymore.

moeyensj:
I've also tested this a few times using fresh Python kernels. I could maybe add a test that runs before all the other unit tests so that we are sure this is fixed but that might be more effort than it is worth.

PR #90, opened 2023-10-25
Add ray as a parallel backend in propagator.py

moeyensj:
Initial attempt at using ray as a parallelization tool for propagation and ephemeris generation. 

No tests have been added, and unit tests will likely fail.

### DIFF
--- a/adam_core/propagator/propagator.py
+++ b/adam_core/propagator/propagator.py
@@ -1,7 +1,7 @@
 import concurrent.futures
 import logging
 from abc import ABC, abstractmethod
-from typing import List, Literal, Optional, Union
+from typing import List, Literal, Optional, Type, Union
 
 import quivr as qv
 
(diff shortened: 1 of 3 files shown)


adam_core/propagator/propagator.py
```diff
@@ -12,26 +15,71 @@
 from ..time import Timestamp
 from .utils import _iterate_chunks
 
+os.environ["OMP_NUM_THREADS"] = "1"
```
akoumjian, line 18: what's all this about?

adam_core/propagator/propagator.py
```diff
@@ -105,30 +158,104 @@ def propagate_orbits(
...
+                if RAY_INSTALLED is False:
+                    raise ImportError(
+                        "Ray must be installed to use the ray parallel backend"
+                    )
+
+                if not ray.is_initialized():
+                    ray.init(num_cpus=max_processes)
+
+                # Add orbits and times to object store
+                times_ref = ray.put(times)
```
akoumjian, line 218: If there is any chance that orbits or times will need to be passed into future runs, the calling function (eg thor) will probably want to send in the refs rather than the literal values so that they don't get duplicated in memory.

I think it's likely we are going to have an unpleasant propagator signature here where we have this odd matrix of choices between parallel backend and propagator, worker functions, sending refs vs values, etc.

All that said, maybe it's easiest to allow a `orbits: Union[Orbits, ObjectRef]` here for now.

akoumjian (approved):
I think it's likely we will want to stick to just one parallel backend once we've had a chance to evaluate the options. I'm not sure if installing `ray` is considered uncomfortable for the casual user, so maybe having it test conditionally for installation is okay.

 Allowing refs to be sent from the upper level I think will be a requirement for us, but other than that if we can get a unit test or two I say we roll with it.

adam_core/propagator/propagator.py
moeyensj: I noticed each ray worker was using 213 threads so this was an attempt to prevent that. Didn't work and I'll remove it before this gets merged.

adam_core/propagator/propagator.py
moeyensj: I've explicitly added support for passing object references. Note that we do need to dereference orbits when we propagate covariances but otherwise the reference just gets passed along. 

Unit tests have also been added to test pass-by reference.

moeyensj:
For the time being, I've just added `ray` as a testing dependency. Once we feel like we are comfortable with using it then we should move to the main dependency list.

PR #98, opened 2023-11-22
Assume missing off-diagonal covariance values are zero

moeyensj:
In some cases, we might have covariance matrices that have only diagonal elements defined with the off-diagonal elements missing (represented as NaNs). Some input coordinates may represent datasets that do not explicitly calculate off-diagonal (covariance) values, but we still want to be able to calculate chi2 for those coordinates using the diagonal (variance) values. In these instances, this PR modifies the residuals code to assume those missing values are zero and emits a warning when this occurs.

### DIFF
--- a/adam_core/coordinates/residuals.py
+++ b/adam_core/coordinates/residuals.py
@@ -77,9 +77,22 @@ def apply_cosine_latitude_correction(
     residuals[:, 1] *= cos_lat
     residuals[:, 4] *= cos_lat
 
-    # Apply the cos(latitude) factor to the covariance matrix
-    covariances = cos_lat_cov @ covariances @ cos_lat_cov.transpose(0, 2, 1)
+    # Identify locations where the covariance matrix is NaN and set
+    # those values to 0.0
+    nan_cov = np.isnan(covariances)
+    covariances_masked = np.where(nan_cov, 0.0, covariances)
 
+    # Apply the cos(latitude) factor to the covariance matrix.
+    covariances_masked = (
+        cos_lat_cov @ covariances_masked @ cos_lat_cov.transpose(0, 2, 1)
+    )
+
+    # Set the covariance matrix to nan where it was nan before.
+    # Note that when we apply cos(latitude) to the covariance matrix in the previous statement,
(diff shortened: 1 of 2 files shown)


adam_core/coordinates/residuals.py
```diff
@@ -310,8 +314,23 @@ def calculate_chi2(
     [1] Carpino, M. et al. (2003). Error statistics of asteroid optical astrometric observations.
         Icarus, 166, 248-270. https://doi.org/10.1016/S0019-1035(03)00051-4
     """
+    # Raise error if any of the diagonal elements are nan
+    if np.any(np.isnan(np.diagonal(covariances, axis1=1, axis2=2))):
+        raise ValueError("Covariance matrix has NaNs on the diagonal.")
+
+    # Warn if any of the non-diagonal elements are nan
+    print(covariances)
```
akoumjian, line 322:
```suggestion
```

adam_core/coordinates/residuals.py
```diff
@@ -310,8 +314,23 @@ def calculate_chi2(
...
+    if np.any(np.isnan(covariances)):
+        warnings.warn(
+            "Covariance matrix has NaNs on the off-diagonal (these will be assumed to be 0.0).",
+            UserWarning,
+        )
+        covariances = np.where(np.isnan(covariances), 0.0, covariances)
+        print(covariances)
```
akoumjian, line 329:
```suggestion
```

adam_core/coordinates/residuals.py
```diff
@@ -310,8 +314,23 @@ def calculate_chi2(
...
+
     W = np.linalg.inv(covariances)
     chi2 = np.einsum("ij,ji->i", np.einsum("ij,ijk->ik", residuals, W), residuals.T)
+    print(chi2)
```
akoumjian, line 333:
```suggestion
```

akoumjian (approved):
Looks good.

adam_core/coordinates/residuals.py
moeyensj: I fixed all of these in the rebase.

PR #99, opened 2023-12-18
Fix orbital period property

moeyensj:
I noticed a bug in the orbital period property but couldn't figure out what was going on. I've added a bunch of JAX-jitted functions for repeated calculations with unit tests and then implemented those in the Keplerian and Cometary coordinate properties. Turns out the deleter for the orbital period was incorrectly defined.


akoumjian (approved):
Good coverage. Looking forward to this fix.

PR #101, opened 2024-01-03
Add fit_least_squares

moeyensj:
Adds a differential correction function that uses `scipy.optimize.least_squares` to perform a least squares fit of an orbit to observations. Also adds a function to remove the worst outlier, however, this PR does not add any logic or procedures to iteratively handle outlier rejection. This is best implemented by external users of these lower-level functions.


adam_core/orbit_determination/differential_correction.py
```diff
@@ -0,0 +1,211 @@
...
+
+
+def fit_least_squares(
+    orbit: Orbits,
+    observations: OrbitDeterminationObservations,
+    propagator: Propagator,
+    **kwargs
+) -> Tuple[FittedOrbits, FittedOrbitMembers]:
+    """
+    Diffentially correct a fitted orbit using least squares.
```
akoumjian, line 88:
```suggestion
    Differentially correct a fitted orbit using least squares.
```

adam_core/orbit_determination/differential_correction.py
```diff
@@ -0,0 +1,211 @@
...
+    fitted_orbit : `~adam_core.orbit_determination.FittedOrbits` (1)
+        Fitted orbit.
+    fitted_orbit_members : `~adam_core.orbit_determination.FittedOrbitMembers` (N)
+        Fitted orbit members.
+    """
+    assert len(orbit) == 1, "Only one orbit can be differentially corrected"
+
+    # TODO: Investigate whether we want to add fitting for the epoch as well
+    # Set up least squares problem
+    func = residual_function
```
akoumjian, line 119: Why rename this if it's not parameterized?

akoumjian (approved):

adam_core/orbit_determination/differential_correction.py
moeyensj: I was toying around with a second version of the function that fits the epoch as well. I'll remove this.

moeyensj: I was toying around with a second version of the function that fits the epoch as well. I'll remove this.

PR #107, opened 2024-04-08
Add tests for calc_orbit_class

moeyensj:


akoumjian:
Would we want this as a method on the quivr table?

moeyensj:
> Would we want this as a method on the quivr table?

Done!

PR #109, opened 2024-05-06
Add ADES utility

moeyensj:
The included `sample_ades.psv` (which the ADES utility is tested against) passes the Minor Planet Center's ADES submission test:
```
curl https://minorplanetcenter.net/submit_psv_test -F "ack=TEST" -F "ac2=email@email.com" -F "source=<sample_ades.psv"
Submission format valid.
```


adam_core/observations/ades.py
```diff
@@ -0,0 +1,291 @@
...
+                    if col in ades.columns:
+                        ades[col] = [
+                            f"{i:.{prec_col}f}"
+                            if i is not None or not np.isnan(i)
+                            else ""
+                            for i in ades[col]
+                        ]
+
+                # This will append to the file since the file is opened in append mode
+                ades.to_csv(f, sep="|", header=True, index=False, float_format="%.16f")
```
akoumjian, line 289: Any particular reason to write to a file directly rather than to a string? Might make it easier to not have to use file objects for testing and such.

adam_core/observations/tests/test_ades.py
```diff
@@ -0,0 +1,235 @@
...
+
+    fundingSource = "WilliamBowes, McGregorGirand, Tito's Vodka, PRawls, SKraus, Yishan/KWong, SGalitsky, Google"
+    comments = [
+        "THIS IS A TEST FILE CONTAINING FAKE OBSERVATIONS",
+        "Discovery candidates found by members of the Asteroid Institute, a program of B612 Foundation,",
+        "using the institute's ADAM::THOR discovery service running on Google Cloud.",
+    ]
+
+    obs_contexts = {
+        "W84": ObsContext(
```
akoumjian, line 49: Instead of just for tests, should we also have some mini fixtures of the ObsContexts to have on hand for datasets we use regularly?

akoumjian (approved):
Just a few questions and comments.

adam_core/observations/ades.py
```diff
@@ -0,0 +1,291 @@
...
+                )
+                ades.insert(
+                    4,
+                    "obsTime",
+                    np.array([i + "Z" for i in observation_times.utc.isot]),
+                )
+                ades.drop(columns=["obsTime.days", "obsTime.nanos"], inplace=True)
+
+                # Multiply rmsRA by cos(dec) since ADES wants the random component in rmsRAcosDec
+                ades.loc[:, "rmsRA"] *= np.cos(np.radians(ades["dec"]))
```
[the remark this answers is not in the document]
moeyensj: That is what it should be doing. Maybe it's better if I make it more explicit:
```python
ades["rmsRA"] = ades["rmsRA"].values * np.cos(np.radians(ades["dec"].values))
```

adam_core/observations/ades.py
```diff
@@ -0,0 +1,291 @@
...
+    rmsDec = qv.Float64Column(nullable=True)
+    mag = qv.Float64Column(nullable=True)
+    rmsMag = qv.Float64Column(nullable=True)
+    band = qv.LargeStringColumn(nullable=True)
+    stn = qv.LargeStringColumn()
+    mode = qv.LargeStringColumn()
+    astCat = qv.LargeStringColumn()
+    remarks = qv.LargeStringColumn(nullable=True)
+
+    def to_psv(
```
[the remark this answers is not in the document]
moeyensj: I'll swap it to a separate function: `write_ADES_to_psv`.

adam_core/observations/tests/test_ades.py
moeyensj: Would it make sense to store them in a separate repo instead of in adam_core (maybe a private `adam-metadata`)?

adam_core/observations/ades.py
moeyensj: Hmm, good point! I think I'd need to write a method to convert the dataframe to a string with the appropriate separator included. It doesn't appear that df.to_string() allows you to define the separator. I'd definitely prefer just working with a string and then having separate function write out a file with that string.

adam_core/observations/tests/test_ades.py
akoumjian: I can understand wanting that separation. Not sure if the overhead is worth it, I could go either way.

akoumjian:
:shipit:

PR #113, opened 2024-05-16
Move PYOORB to a namespace package and update packaging utilities

akoumjian:


README.md
```diff
@@ -150,8 +150,9 @@ spherical_elements = orbits.coordinates.to_spherical()
 ```
 
 ### Propagator
-The propagator class in `adam_core` provides a generalized interface to the supported orbit integrators and ephemeris generators. By default,
-`adam_core` ships with PYOORB.
+The propagator class in `adam_core` provides a generalized interface to the supported orbit integrators and ephemeris generators. The propagator class is designed to be used with the `Orbits` class and can handle multiple orbits and times. 
+
+You will need to install either adam_core[pyoorb] or adam_core[assist], or another compatible propagator in order to use propagation, ephemeris generation, or impact analysis.
```
moeyensj, line 155: Do we want to add `adam_core[all]` and/or `adam_core[default]`?

pyproject.toml
```diff
@@ -1,9 +1,6 @@
-[build-system]
-requires = ["hatchling"]
-build-backend = "hatchling.build"
-
 [project]
 name = "adam_core"
+version = "0.1.0"
```
moeyensj, line 3: Does this mean we remove the semantic git-hash based versioning? It might have already been removed with the hatch stuff. Just curious.

moeyensj (approved):
Looks good to me. No major comments. It's always nice to see the number of lines deleted greatly exceed the number of lines added.

README.md
akoumjian: Yeah, I'm going to actually implement these now.

pyproject.toml
akoumjian: I can potentially put it back. I was running into issues.

PR #115, opened 2024-07-26
Add generic Ephemeris implementation

akoumjian:
This PR uses functions generated by @moeyensj and @KatKiker to allow any adam-core propagator the ability to generate lighttime corrected Ephemeris.

### DIFF
--- a/src/adam_core/orbits/query/horizons.py
+++ b/src/adam_core/orbits/query/horizons.py
@@ -226,9 +226,6 @@ def query_horizons_ephemeris(
         ignore_index=True,
     )
 
-    # Horizons produces UTC but we use tdb everywhere
-    epochs = Timestamp.from_jd(pa.array(dfs["datetime_jd"]), scale="utc").rescale("tdb")
-
     ephemeris = Ephemeris.from_kwargs(
         orbit_id=dfs["orbit_id"],
         object_id=dfs["targetname"],
@@ -236,7 +233,7 @@ def query_horizons_ephemeris(
         light_time=dfs["lighttime"] / 1440,
         alpha=dfs["alpha"],
         coordinates=SphericalCoordinates.from_kwargs(
-            time=epochs,
+            time=Timestamp.from_jd(pa.array(dfs["datetime_jd"]), scale="utc"),
             lon=dfs["RA"],
             lat=dfs["DEC"],
             origin=Origin.from_kwargs(code=dfs["observatory_code"]),
--- a/src/adam_core/propagator/propagator.py
+++ b/src/adam_core/propagator/propagator.py
@@ -220,6 +220,12 @@ class EphemerisMixin:
                 spherical_coordinates, SphericalCoordinates, frame_out="equatorial"
             )
 
+            # Ephemeris are generally compared in UTC, so rescale the time
+            spherical_coordinates = spherical_coordinates.set_column(
+                "time",
+                spherical_coordinates.time.rescale("utc"),
+            )
+
             if isinstance(orbits, Orbits):
 
                 ephemeris = Ephemeris.from_kwargs(


src/adam_core/orbits/query/horizons.py
```diff
@@ -208,19 +212,38 @@ def query_horizons_ephemeris(
...
-    ephemeris.sort_values(
+    dfs = pd.concat(dfs, ignore_index=True)
+    dfs.sort_values(
         by=["orbit_id", "datetime_jd", "observatory_code"],
         inplace=True,
         ignore_index=True,
     )
+
+    # Horizons produces UTC but we use tdb everywhere
+    epochs = Timestamp.from_jd(pa.array(dfs["datetime_jd"]), scale="utc").rescale("tdb")
```
moeyensj, line 230: When we use ephemeris (like from the propagator class), we typically compare them to observations which are more likely to be in UTC so I think it makes sense to leave the ephemeris in UTC and have the user rescale downstream if needed. That said though, we haven't really used these specific functions for anything but testing so maybe its not as much of an issue.

src/adam_core/propagator/propagator.py
```diff
@@ -6,12 +6,16 @@
 import numpy.typing as npt
 import quivr as qv
 
-from adam_core.ray_cluster import initialize_use_ray
-
+from ..coordinates.cartesian import CartesianCoordinates
+from ..coordinates.origin import Origin, OriginCodes
+from ..coordinates.spherical import SphericalCoordinates
+from ..coordinates.transform import transform_coordinates
+from ..dynamics.aberrations import C
```
moeyensj, line 13:
```suggestion
from ..constants import Constants as c

C = c.C
```

Might be a bit cleaner to import the constant from where it's defined.

moeyensj: My suggestion will fail linting.

src/adam_core/propagator/propagator.py
```diff
@@ -89,17 +93,158 @@ class EphemerisMixin:
     Subclasses should implement the _generate_ephemeris method.
     """
 
-    @abstractmethod
+    def _add_light_time(
+        self,
+        orbits,
+        observers,
+        lt_tol: float = 1e-12,
+        max_iter: int = 10,
```
moeyensj, line 101: I had set this to 10 just because its a nice number but feel free to change this. Our examples from yesterday seemed to complete in about 4 iterations.

src/adam_core/propagator/propagator.py
```diff
@@ -89,17 +93,158 @@ class EphemerisMixin:
...
+                - observers_barycentric.coordinates.values[:, 4],
+                vz=propagated_orbits_aberrated.coordinates.values[:, 5]
+                - observers_barycentric.coordinates.values[:, 5],
+                covariance=None,
+                # The ephemeris times are at the point of the observer,
+                # not the aberated orbit
+                time=observers.coordinates.time,
+                origin=Origin.from_kwargs(code=observer_codes),
+                frame="ecliptic",
+            )
```
moeyensj, line 211:
```suggestion
            topocentric_state = propagated_orbits_aberrated.coordinates.values - observers_barycentric.coordinates.values
            topocentric_coordinates = CartesianCoordinates.from_kwargs(
                x=topocentric_state[:, 0],
                y=topocentric_state[:, 1],
                z=topocentric_state[:, 2],
                vx=topocentric_state[:, 3],
                vy=topocentric_state[:, 4],
                vz=topocentric_state[:, 5],
                covariance=None,
                # The ephemeris times are at the point of the observer,
                # not the aberrated orbit
                time=observers.coordinates.time,
                origin=Origin.from_kwargs(code=observer_codes),
                frame="ecliptic",
            )
```

Probably will only get a small performance gain but this way we only call `np.array(self.table.select(["x", "y", "z", "vx", "vy", "vz"]))` once.

moeyensj: Maybe one day we overload the math operators for our coordinates classes.

moeyensj (approved):
Exciting!

src/adam_core/orbits/query/horizons.py
akoumjian: Maybe it makes more sense for our generic ephemeris to return in UTC then.

src/adam_core/propagator/propagator.py
moeyensj: > Probably will only get a small performance gain but this way we only call np.array(self.table.select(["x", "y", "z", "vx", "vy", "vz"])) once.
Twice not once but at least not 12 times!

src/adam_core/propagator/propagator.py
akoumjian: I don't really have a sense of the long tail curve for convergence (if there even is one). In my mind, it's perfectly okay if the max_iter is literally never reached, it's only there to keep something that won't converge from running on forever.

src/adam_core/orbits/query/horizons.py
akoumjian: Updated this to return UTC (as well as our generic ephemeris method).

PR #116, opened 2024-08-12
Return propagated orbits with original origins and frame based on inp…

akoumjian:
…ut orbit_id

This updates the high level `.propagate_orbits` method to return orbits with their original `Origin` and `frame`. Now this logic will hold for all pluggable backends for the `Propagator` class.

### DIFF
--- a/src/adam_core/dynamics/impacts.py
+++ b/src/adam_core/dynamics/impacts.py
@@ -1,4 +1,3 @@
-import importlib.util
 import logging
 from abc import abstractmethod
 from typing import List, Optional, Tuple
--- a/src/adam_core/dynamics/impacts.py
+++ b/src/adam_core/dynamics/impacts.py
@@ -21,13 +21,6 @@ from ..time import Timestamp
 
 logger = logging.getLogger(__name__)
 
-# Test to see that at least one impact-enabled propagator is
-# installed and if not print a warning
-if importlib.util.find_spec("adam_core.propagator.adam_assist") is None:
-    logger.warning(
-        "No impact-enabled propagator installed. Impact calculations will not be possible."
-    )
-
 RAY_INSTALLED = False
 try:
     import ray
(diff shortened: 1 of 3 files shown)


ntellis (approved):
Looks good to me, I have (minimally) tested it locally.

src/adam_core/dynamics/impacts.py
```diff
@@ -21,13 +20,6 @@
 
 logger = logging.getLogger(__name__)
 
-# Test to see that at least one impact-enabled propagator is
-# installed and if not print a warning
-if importlib.util.find_spec("adam_core.propagator.adam_assist") is None:
-    logger.warning(
-        "No impact-enabled propagator installed. Impact calculations will not be possible."
-    )
-
```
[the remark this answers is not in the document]
akoumjian: No, I decided it was better to simply error out when trying to run a propagator without the impact mixin.

src/adam_core/propagator/propagator.py
```diff
@@ -575,6 +577,29 @@ def propagate_orbits(
...
+        # Return the results with the original origin and frame
+        # Preserve the original output origin for the input orbits
+        # by orbit id
+        final_results = None
+        unique_origins = pc.unique(orbits.coordinates.origin.code)
+        for origin_code in unique_origins:
+            origin_orbits = orbits.select("coordinates.origin.code", origin_code)
+            result_origin_orbits = propagated.where(
+                pc.field("orbit_id").isin(origin_orbits.orbit_id)
+            )
```
moeyensj, line 589: Not a blocker, is this comparable in speed?
```
result_origin_orbits = propagated.apply_mask(pc.is_in(propagated.orbit_id, origin_orbits.orbit_id))
```

moeyensj (approved):

src/adam_core/propagator/propagator.py
akoumjian: I hadn't even considered these would be different. Have you seen differences between using `pc.field` and passing the column directly?

akoumjian: It does look like your version is slightly faster. I filtered 100k orbits against 20k orbit ids. 
```
Existing: 0.00540
New:      0.00497
```

Pushing an update now.

PR #117, opened 2024-09-17
Upgrade to numpy>=2.0.0

moeyensj:


.github/workflows/pip-build-lint-test-coverage.yml
```diff
@@ -26,21 +26,11 @@ jobs:
...
-          wget  -P /tmp/ https://storage.googleapis.com/oorb-data/oorb_data.tar.gz
-          mkdir -p /tmp/oorb/
-          tar -xvf /tmp/oorb_data.tar.gz -C /tmp/oorb/      
+          cache: 'pip' # caching pip dependencies 
       - name: Install Testing Dependencies
         run: |
           pip install pip --upgrade
           pip install ".[dev]"
-          # Install pyoorb propagator from source
-          pip install git+https://github.com/B612-Asteroid-Institute/adam-pyoorb.git@main
```
moeyensj, line 43: Controversial maybe but I think it is time.

akoumjian:
I tested adam-assist against this version and unit and integration tests all passed. I was worried that maybe the serialization to numpy arrays in rebound might behave differently, since I think the C apis have changed, but as far as I can tell it's fine.

Looks like unit tests for adam-core pass in about same amount of time (~1min 30s), so not obvious performance from that alone. adam_etl might have the biggest changes, but we were due for a big upgrade anyway.

akoumjian (approved):

PR #118, opened 2024-09-17
Set static_argnames for relevant jax-jitted functions

moeyensj:


akoumjian:
Maybe merge in the numpy 2 changes and see if that issue resolves itself.

PR #119, opened 2024-09-18
Add SourceCatalog

moeyensj:
We've been discussing consolidating our input type definitions across ADAM-ETL, THOR, the impact project, etc.. This PR adds a de-normalized and flattened input type called `SourceCatalog` with methods to convert subsets of its columns into the more compute-oriented types. 

I think we will also want to add an "Observations" type that stores the relevant subset columns in their appropriate SubTable types. So something like:

```python

class Observations(qv.Table):
    id = qv.LargeStringColumn()
    coordinates = SphericalCoordinates.as_column()
    observers = Observers.as_column()
    ... etc

    def from_source_catalog(cls, source_catalog):
         ...
```

This is still a WIP.


akoumjian (commented):
I looked at precovery indexing and this should work fine for it.
Do we want a method to get the healpixel at a certain resolution? Or does that better belong on `SphericalCoordinates`?

moeyensj:
I think attaching it to SourceCatalog is totally fine.

akoumjian (approved):

PR #120, opened 2024-10-01
Add fix for non positive semi definite matrices 

KatKiker:


src/adam_core/coordinates/covariances.py
```diff
@@ -211,7 +237,23 @@ def sample_covariance_random(
         Weights of the samples.
     W_cov: `~numpy.ndarray` (num_samples)
         Weights of the samples to reconstruct covariance matrix.
+
+    Raises
+    ------
+    ValueError : If the covariance matrix is not positive semidefinite, within the given tolerance.
+
     """
+    print(np.linalg.eigvals(cov))
```
akoumjian, line 246:
```suggestion
```

src/adam_core/coordinates/covariances.py
```diff
@@ -179,11 +179,37 @@ def is_all_nan(self) -> bool:
...
+    cov_psd = eigenvectors @ np.diag(eigenvalues) @ eigenvectors.T
+    return cov_psd
+
+
 def sample_covariance_random(
     mean: np.ndarray,
     cov: np.ndarray,
     num_samples: int = 10000,
     seed: Optional[int] = None,
+    tol: Optional[float] = 1e-15,
```
akoumjian, line 212: Let's use a more explicit name, like `semidef_tol`

src/adam_core/coordinates/tests/test_covariances.py
```diff
@@ -44,9 +45,25 @@ def test_sample_covariance_sigma_points():
...
+
+    psd_matrix = make_positive_semidefinite(non_psd_matrix, 1e-10)
+
+    assert np.all(np.linalg.eigvals(psd_matrix) >= 0)
+
+
 def test_sample_covariance_random():
-    # Get a sample of real orbits and test that random sampling
-    # allows the state vector and its covariance to be reconstructed
+    # Get a sample of real orbits    # allows the state vector and its covariance to be reconstructed
```
akoumjian, line 66:
```suggestion
    # Get a sample of real orbits
    # allows the state vector and its covariance to be reconstructed
```

src/adam_core/coordinates/tests/test_covariances.py
```diff
@@ -44,9 +45,25 @@ def test_sample_covariance_sigma_points():
         np.testing.assert_allclose(covariance_sg, covariance, rtol=0, atol=1e-14)
 
 
+def test_make_positive_semidefinite():
```
akoumjian, line 48: Can we get a test of `sample_covariance_random` that verifies the tolerance is working as expected? One where it fixes it within the tolerance and one that throws the error since it's outside the tolerance?

akoumjian (changes requested):
Couple small changes.

akoumjian (approved):

PR #122, opened 2024-10-22
Query scout and get variant orbits back

akoumjian:


src/adam_core/orbits/query/scout.py
```diff
@@ -0,0 +1,211 @@
...
+    variant_orbits : `~adam_core.orbits.VariantOrbits`
+        Table containing the variant orbits
+    """
+    cometary_coords = CometaryCoordinates.from_kwargs(
+        q=pc.cast(scout_orbits.qr, pa.float64()),
+        e=pc.cast(scout_orbits.ec, pa.float64()),
+        i=pc.cast(scout_orbits.inc, pa.float64()),
+        raan=pc.cast(scout_orbits.om, pa.float64()),
+        ap=pc.cast(scout_orbits.w, pa.float64()),
+        tp=pc.cast(scout_orbits.tp, pa.float64()),
```
moeyensj, line 159:
```suggestion
        tp=pc.subtract(pc.cast(scout_orbits.tp, pa.float64()), 2400000.5),
```

src/adam_core/orbits/variants.py
```diff
@@ -168,6 +168,57 @@ def collapse(self, orbits: Orbits) -> Orbits:
...
+            mean = np.average(
+                object_variants.coordinates.values,
+                axis=0,
+            )
+
+            # Calculate the covariance matrix
+            covariance = weighted_covariance(
+                mean,
+                object_variants.coordinates.values,
+                np.ones(len(object_variants), dtype=np.float64),
```
moeyensj, line 200: I think weights should sum to 1 but since its a constant value, I'm not sure it will have an impact on the result.

```suggestion
                np.ones(len(object_variants), dtype=np.float64) / len(object_variants),
```

src/adam_core/orbits/variants.py
```diff
@@ -168,6 +168,57 @@ def collapse(self, orbits: Orbits) -> Orbits:
...
+                object_id=[object_id],
+                coordinates=CartesianCoordinates.from_kwargs(
+                    x=[mean[0]],
+                    y=[mean[1]],
+                    z=[mean[2]],
+                    vx=[mean[3]],
+                    vy=[mean[4]],
+                    vz=[mean[5]],
+                    covariance=CoordinateCovariances.from_matrix(covariance),
+                    time=object_variants.coordinates.time[0],
```
moeyensj, line 215: Are they all defined at the same time?

moeyensj: Might be good to throw in an assert to check that they are.

moeyensj (changes requested):

PR #125, opened 2024-11-14
Port iod/od from THOR to adam_core

ntellis:
This adds TIOR iod/od to adam_core, and adds some _very basic_ unit tests to cover them. Most subcomponents (gauss etc) don't have their own tests.

I need to switch out the way that assist is being imported in tests, obviously (since it's a hardcoded path at the moment).


src/adam_core/orbit_determination/tests/test_iod.py
```diff
@@ -0,0 +1,108 @@
...
+    spec = importlib.util.spec_from_file_location(
+        "adam_core.propagator.adam_assist", assist_path
+    )
+    adam_assist = importlib.util.module_from_spec(spec)
+    sys.modules["adam_core.propagator.adam_assist"] = adam_assist
+    spec.loader.exec_module(adam_assist)
+    from adam_core.propagator.adam_assist import ASSISTPropagator  # noqa: E402
+except Exception as e:
+    ASSISTPropagator = None
+    print(e)
```
akoumjian, line 36: Before we merge this in, I'd like to see if we can get this to work by adding `adam-assist` as an editable dependency using the git repo.

https://pdm-project.org/latest/usage/dependency/#editable-dependencies

This will require changing the way we do test / dev dependencies, away from `optional-dependencies` and instead using the pdm specific method.

https://pdm-project.org/latest/usage/dependency/#add-development-only-dependencies

akoumjian (approved):

PR #141, opened 2025-01-30
Add query_neocc

moeyensj:


src/adam_core/orbits/query/neocc.py
```diff
@@ -0,0 +1,260 @@
...
+                                6,
+                            )
+                        ),
+                        frame="ecliptic",
+                        origin=Origin.from_kwargs(code=["SUN"]),
+                    ).to_cartesian(),
+                )
+                orbits = qv.concatenate([orbits, orbit])
+
+        except requests.RequestException as e:
```
akoumjian, line 256: I don't think we should capture the exception here. Let the caller decide what they want to do with a failed query. It will be more obvious than forgetting to handle the None and looking at the logs.

akoumjian (changes requested):
I would allow the request exception to percolate and then just the formatting.

moeyensj:
I think this is ready to be squashed and merged now.

PR #148, opened 2025-03-05
Add calculate_perturber_moids

moeyensj:


src/adam_core/dynamics/moid.py
```diff
@@ -21,14 +21,34 @@
...
-from ..coordinates import CartesianCoordinates, KeplerianCoordinates
+from ..coordinates import (
+    CartesianCoordinates,
+    KeplerianCoordinates,
+    Origin,
+    OriginCodes,
+)
 from ..dynamics.propagation import _propagate_2body
 from ..orbits import Orbits
+from ..propagator.utils import _iterate_chunks
```
akoumjian, line 41: We should move this to some kind of iter utils module

src/adam_core/dynamics/moid.py
```diff
@@ -145,13 +165,129 @@ def calculate_moid(
...
+
+
+moid_worker_ray = ray.remote(moid_worker)
+
+
+def calculate_perturber_moids(
+    orbits: Orbits,
+    perturber: Union[OriginCodes, List[OriginCodes]],
+    chunk_size: int = 100,
+    max_processes: int = 1,
```
akoumjian, line 219:
```suggestion
    max_processes: Optional[int] = 1,
```

src/adam_core/dynamics/moid.py
```diff
@@ -145,13 +165,129 @@ def calculate_moid(
...
+    max_processes : int, optional
+        The maximum number of processes to use, by default 1.
+
+    Returns
+    -------
+    PerturberMOIDs
+        A table containing the MOID and time for each orbit with respect to the perturbing body or bodies.
+    """
+    assert len(orbits.orbit_id.unique()) == len(orbits)
+    assert len(orbits.coordinates.origin.code.unique()) == 1
```
akoumjian, line 242: Probably something obvious, but why this assert?

akoumjian (approved):

src/adam_core/dynamics/moid.py
moeyensj: If we wanted to accept arbitrary origins then we would need to batch by the orbits' unique origin as well (since we need to pull the perturber states to have the same origin). I suppose this MOID calculation really only makes sense from the heliocenter or barycenter at the moment.

src/adam_core/dynamics/moid.py
moeyensj: Done.

PR #152, opened 2025-05-07
Prototype Observer and Ephemeris generation for space telescopes usig…

akoumjian:
…n SPICE kernels

AI Generated Summary

#Space Telescopes Observer Support via SPICE Kernels

This PR adds support for space telescopes as observers in the adam-core library using SPICE kernels. The implementation provides a flexible framework for registering and using custom SPICE kernels to obtain state vectors for space-based observatories.

## Key Changes:

1. **Space Telescope Observer Support**: Added ability to use space-based telescopes like JWST as observers by integrating SPICE kernel support for retrieving telescope ephemerides.

2. **Observer Code Refactoring**: Refactored the observer state code to better separate Earth-based (MPC) observers from space-based observers, improving code organization and maintainability.

3. **SPICE Kernel Management**: 
   - Added new functions to register, unregister, and list custom SPICE kernels
   - Implemented kernel caching to track loaded kernels and prevent duplicate loading

4. **New Utility Functions**:
   - `get_spice_body_state()`: Gets state vectors for any SPICE body
   - `register_spice_kernel()` / `unregister_spice_kernel()`: Manage custom kernels
   - `list_registered_kernels()`: Track loaded kernels 

5. **Testing Infrastructure**:
   - Added test module for JWST observer with test cases for SPICE kernel integration
   - Added test module for generating ephemeris using space-based observers
   - Added JWST SPICE kernel (covering 2020-2024) for testing purposes
   - Test cases for mixed observer types (space and ground-based)

6. **Observer Precedence**:
   - Implemented priority system favoring MPC observatory codes over SPICE body codes when both exist

This implementation enables observation and ephemeris generation from space-based platforms, expanding the library's capabilities for a wider range of use cases including space telescope observation simulation and planning.


src/adam_core/observers/state.py
```diff
@@ -8,14 +8,127 @@
...
+        The MPC observatory code of the observer.
+    times : Timestamp
+        The times at which to get the observer state.
+    frame : {'ecliptic', 'equatorial', 'itrf93'}
+        The frame in which to return the observer state.
+    """
+    if frame == "ecliptic":
+        frame_spice = "ECLIPJ2000"
+    elif frame == "equatorial":
+        frame_spice = "J2000"
```
moeyensj, line 42: Are we missing an IRTF93 check here?

moeyensj: Would actually be a funny frame to request an Earth-surface observer state in because it should be nearly constant. I could see it being useful for a satellite or spacecraft though.

src/adam_core/observers/state.py
```diff
@@ -8,14 +8,127 @@
...
+    observer_states = CartesianCoordinates.from_kwargs(
+        time=times,
+        x=r_obs[:, 0],
+        y=r_obs[:, 1],
+        z=r_obs[:, 2],
+        vx=v_obs[:, 0],
+        vy=v_obs[:, 1],
+        vz=v_obs[:, 2],
+        frame=frame,
+        origin=Origin.from_kwargs(code=[origin.name for i in range(len(times))]),
```
moeyensj, line 126: Nitpick: maybe a `pa.repeat` to avoid a list comprehension.

src/adam_core/observers/state.py
```diff
@@ -74,92 +178,17 @@ def get_observer_state(
...
-            origin=Origin.from_kwargs(code=[origin.name for i in range(len(times))]),
+    # Try to retrieve the body ID from SPICE, could be a custom SPICE kernel NAIF code
+    try:
+        body_id = sp.bodn2c(code)
+    except sp.SpiceyError:
+        err = f"{code} is not a valid MPC observatory code and was not found in SPICE kernels."
+        raise ValueError(err)
+    else:
+        return get_spice_body_state(
+            body_id=body_id, times=times, frame=frame, origin=origin
```
moeyensj, line 193: Great way to handle this.

moeyensj (commented):
This is pretty clean. How do we handle MPC observatory codes for spacecraft? Say that I do `Observers.from_code("C51", times)`. I'm not sure we should but I can see someone trying something like that.

src/adam_core/observers/state.py
akoumjian: I guess we could. What would be the "correct" way to do this?

akoumjian: I guess you would get the rotational velocity would could be useful, yeah?

moeyensj: Yeah that is true, you'd get the rotational velocity. I think as implemented it ought to work.

akoumjian: Okay, I've updated with a reviewed AI suggestion. I think it looks correct, though maybe it makes sense to grab earth state higher up to make it less verbose.

PR #158, opened 2025-06-03
OpenSpace and ADES improvements for Rubin

moeyensj:


moeyensj:
I think this is ready for the first review round. I still need to do a bit more testing of the SPICE-based OpenSpace visuals but the other two are working correctly.

akoumjian:
@moeyensj Made a couple small changes and fixed an ADES test to match the new slightly larger default precision numbers. When would you need this released to PyPI or conda-forge?

PR #175, opened 2025-11-05
Speed up Timestamp rescale

HoodedCrow:
Create new rescale method in Timestamp class (and rename old one to rescale_astropy).
Use simpler calculations for TDB offset, differs from astropy by up to tens of microseconds.
Use direct calls to ERFA for UTC (because of leapseconds) and fold other offsets to avoid multi-step conversions.
Stick with astropy for UT1 conversions, because those require interpolation of external tables.
Use plain arithmetic (no fancy 4-step stuff) and numpy instead of pyarrow for math throughout.
Speedup on the order of 50% compared to astropy implementation with some introduced rounding errors of about 20ns for UTC conversion and 30us for TDB.
Two old time tests needed to be modified because of the rounding errors.
New tests and benchmarks are added to compare the new implementation with astropy.

### DIFF
--- a/src/adam_core/time/tests/test_time.py
+++ b/src/adam_core/time/tests/test_time.py
@@ -574,60 +574,54 @@ def test_Timestamp_link(rescale):
     # Test that the link method works as expected
     time = Timestamp.from_kwargs(
         days=[68000, 68000, 68010, 68010, 68020, 68020],
-        nanos=[1_500, 2_500, 3_500, 3_500, 4_500, 4_500],
+        nanos=[1, 2, 3, 3, 4, 4],
         scale="tdb",
     )
 
     other = Timestamp.from_kwargs(
-        days=[68000, 68000, 68010, 68020],
-        nanos=[1_500, 2_500, 3_500, 4_500],
-        scale="tdb",
+        days=[68000, 68000, 68010, 68020], nanos=[1, 2, 3, 4], scale="tdb"
     )
     if rescale:
         other = other.rescale("utc")
 
-    # utc-tbd conversion adds a rounding error on the orider of 20ns
-    precision = "us" if rescale else "ns"
-    tail = 0 if rescale else 500
-    tol = 30 if rescale else 0
-    linkage = time.link(other, precision)
+    linkage = time.link(other)
     assert len(linkage.all_unique_values) == 4
 
-    key = (68000, 1_000 + tail)
+    key = (68000, 1)
     left_table, right_table = linkage.select(key)
     assert len(left_table) == 1
     assert len(right_table) == 1
     assert left_table.days.to_pylist() == [68000]
-    assert np.allclose(left_table.nanos.to_numpy(), [1_500], atol=tol)
+    assert left_table.nanos.to_pylist() == [1]
     assert right_table.days.to_pylist() == [68000]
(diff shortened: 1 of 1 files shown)


akoumjian:
Thanks @HoodedCrow !

Finally got a chance to look at this. A 50% speedup is fantastic. I'd like to see if we can keep that speed up but try to avoid the 10s of microseconds for TDB offset and even eliminate the rounding errors. For most cases, the microsecond offset won't matter, but with the combination of Rubin level astrometry (1-5 milliarcseconds) and looking at NEAs, light time correction, and impact corridors, it could be significant. 

I understand that part of the gains comes from doing less work (e.g. analytical/periodic approach vs fetching planet ephemeris to get harmonics). I'm wondering if we can explore implementing our own kernels for this and make them even faster than the ERFA implementation. My cursory understanding is that it essentially involves fetching peturber ephemeris, but maybe ERFA is doing something more clever.

Additionally, I'd like to see if we can make some of these conversions using integer math, such as done in the hifitime library (there implementation isn't perfect, but it sets a high bar). We can potentially avoid floating point errors.

ntellis:
It looks great to me! Feels right now like we would want to keep this as a parallel implementation for situations where we value speed over the highest precision - accumulation of rounding errors over time can affect things like a precise risk corridor estimation etc.

I'd be curious to see if this could be sped up much further with `jax` - being mostly numpy already this should be doable. A good example of this is in [add_light_time](https://github.com/B612-Asteroid-Institute/adam_core/blob/3f3e79c0558d0398b3e039f3f7aa435587c0e086/src/adam_core/dynamics/aberrations.py#L112). Important to note that in order to avoid repeated `jax` recompilation we try to run jitted functions in same-size chunks by padding them, ex:
https://github.com/B612-Asteroid-Institute/adam_core/blob/3f3e79c0558d0398b3e039f3f7aa435587c0e086/src/adam_core/propagator/propagator.py#L137-L169

HoodedCrow:
Thanks for the comments! I'd be happy to jump on a call to discuss this if this helps.

First, I have left the original code still there. It's called `rescale_astropy` now. I am using it in tests, but it would be interesting to see if there is any practical delta introduced by the rounding differences.

Rounding difference is most likely due to the simplified arithmetic. After checking with John C, microseconds difference might matter in some cases (specifically, multiple gravity assists), but nanoseconds probably won't make any difference. Simpler arithmetic does give a nice speedup though.

The microseconds difference from AstroPy on TDB is due to the different algorithm. I am using the version I got from ESA website, which is the same HiFiTime is using (and claiming it's the correct one). AstroPy uses a different approach. Now, the question is, where do you apply this? My understanding is that some databases (including some JPL stuff) don't "modernize" algorithms for years, and if you use those databases you might want to be "wrong the same way they are" instead of using a more correct version. So it would depend on your applications.

ERFA is only there because I don't want to deal with the leapseconds. It's possible to reimplement the leap second function, but I didn't want to do it in the same PR. Once we get rid of ERFA, then it would make sense to revisit float vs long int.

I'll poke at jax, thanks for the pointer! But also likely in a separate PR.

Do you folks have some precision-sensitive test scenarios we could use to check if the rounding errors and the TDB delta make a difference?

akoumjian:
@HoodedCrow I've merged it into main after making a couple stylistic updates and expanding the benchmark tests a bit. Thank you!

PR #200, opened 2026-06-09
Rotation period mvp

ntellis:
# Add a rotation-period estimator with measured confidence

Adds `adam_core.photometry.estimate_rotation_period` — recovers an asteroid's rotation
period from sparse multi-band photometry and returns a **categorical, measured confidence
verdict** instead of a bare point estimate.

- **API:** `estimate_rotation_period(observations, *, method_mode="fourier", …) -> RotationPeriodResult`;
  `RotationPeriodObservations` / `RotationPeriodResult` quivr tables, with
  `RotationPeriodObservations.from_point_source_observations(detections, exposures, object_coords)`
  and detection-level wrappers.
- **D1 contract (headline):** every result carries `period_verdict ∈ {single_period,
  period_family, insufficient_data}` + `reliability_code ∈ {3,2,1}`. The verdict is
  **measured, not promised**: the solver is designed to hedge harmonic aliases to
  `period_family` rather than emit a confident `single_period`, and that behaviour is
  *measured* against standard candles. It is **not** a guarantee of zero aliased
  `single_period` calls — there is a known, tracked residual risk (see Validation).
- **Validation:** against LCDB/DAMIT standard candles (≈0.89 strict / ≈0.95 family
  single-period precision). 11 committed fixtures drive a D1 schema gate and a D4
  zero-false-confidence gate. That gate is a **strict `xfail`** on one known residual
  alias: 1627 Ivar is reported `single_period` at a ~5/3 diurnal-sampling alias
  (P_rec 7.99 h vs P_true 4.795 h). It is tracked calibration debt (frequency-aware
  order-selection work); `strict=True` forces removing the marker once the alias is
  fixed. **Bottom line: measured precision with a known, nonzero residual
  false-confidence risk — not a zero-alias guarantee.**
- **Scope:** rotation-period subsystem only; no new hard deps (JAX optional, numpy default).

A supporting internal Greenstreet-76 regression run (a conservative-confidence guard, *not*
a paper reproduction; run on a patched harness with exact reruns disabled) found 0
false-confidence among its 6 `single_period` calls; it does **not** override the committed
1627 Ivar `xfail`.

Validation harness lives in `B612-Asteroid-Institute/adam-core-rotation-validation` (private).


ntellis:
**On the false-confidence risk** (re: the review finding)

To be explicit about what this estimator does and does not promise: the confidence verdict is **measured, not guaranteed.** The solver is designed to hedge harmonic and sampling aliases to `period_family` rather than assert a confident `single_period`, but that is *calibrated* behaviour with a known, nonzero residual risk — a `single_period` result can occasionally be a harmonic or sampling alias of the true period.

There is one tracked case on the committed set: **1627 Ivar**, reported `single_period` at a ~5/3 diurnal-sampling alias (P_rec 7.99 h vs P_true 4.795 h). It is kept as a **strict `xfail`** (calibration debt; the fix is frequency-aware order selection) so that resolving it surfaces as an xpass and forces removing the marker. Measured strict `single_period` precision on the LCDB/DAMIT standard-candle set is ~0.88.

Following the review, this is now documented honestly rather than overclaimed:
- The PR description no longer claims zero aliased `single_period`; it states measured precision with explicit known residual risk.
- The `estimate_rotation_period` docstring has a `Notes` section on this risk, advising a cross-check of any `single_period` result against `alternate_period_days` / `reliability_code` when an integer-factor error would be costly (`0d4bb63b`).

Practical guidance for users: treat `single_period` as high-confidence but not infallible; `period_family` and `insufficient_data` are the honest hedges, not failures.

akoumjian (approved):
Looks good to me.

akoumjian:
@ntellis Please add some narrative docs for using this new functionality, so it can be discovered.

PR #11, opened 2023-02-22
Add adam_core.utils.helpers

moeyensj:
Adds two functions with some test data:

```python
from adam_core.utils.helpers import make_real_orbits
from adam_core.utils.helpers import make_simple_orbits

orbits = make_real_orbits()
orbits.to_df()
orbit_id                        object_id  mjd_tdb          q         e           i  ...      cov_tp_i   cov_tp_raan     cov_tp_ap     cov_tp_tp       origin     frame
0          0  594913 'Aylo'chaxnim (2020 AV2)  59038.0   0.457155  0.176960   15.868082  ... -2.570466e-09 -3.305573e-09  2.960373e-08  1.554928e-08  heliocenter  ecliptic
1          1         163693 Atira (2003 CP20)  57255.0   0.502323  0.322134   25.619293  ...  7.850735e-12 -2.997388e-13  6.341497e-13  2.814448e-11  heliocenter  ecliptic
2          2                       (2010 TK7)  56584.0   0.809319  0.190762   20.884723  ...  9.987907e-11  2.028809e-10  2.241994e-09  3.088908e-09  heliocenter  ecliptic
3          3          3753 Cruithne (1986 TO)  56899.0   0.484019  0.514862   19.807479  ... -1.253603e-11 -6.902988e-11  1.295538e-10  2.117343e-10  heliocenter  ecliptic
4          4            54509 YORP (2000 PH5)  52655.0   0.770117  0.229915    1.833144  ... -2.609772e-11 -3.642007e-10  3.112754e-10  1.569425e-10  heliocenter  ecliptic
5          5           2063 Bacchus (1977 HB)  57210.0   0.701358  0.349362    9.432852  ...  4.939622e-13  1.165165e-12  6.197185e-12  8.265897e-12  heliocenter  ecliptic
6          6             1221 Amor (1932 EA1)  57836.0   1.083156  0.435605   11.879488  ... -3.996187e-12 -6.945640e-11  2.036742e-10  4.351453e-10  heliocenter  ecliptic
7          7               433 Eros (A898 PA)  53311.0   1.133355  0.222808   10.829184  ...  2.293673e-13 -1.020773e-12  4.865456e-12  5.908122e-12  heliocenter  ecliptic
8          8               3908 Nyx (1980 PA)  54128.0   1.043592  0.458543    2.181544  ... -4.883613e-13  1.371504e-12  1.176539e-12  2.931765e-12  heliocenter  ecliptic
[...]

PR #10, opened 2023-02-22
Fix SBDB query handling of responses with single epoch

spenczar:
Lots of test infrastructure too

PR #201, opened 2026-06-10
Nongrav support

KatKiker:
## Summary

Adds canonical non-gravitational force support to `adam_core`, per the team decision to ingest/store Marsden-style accelerations only, with solved-parameter uncertainty folded into the coordinate covariance.

### Model

- `NonGravitationalParameters` stores `source`, the fitted accelerations `A1/A2/A3` (au/d², radial/transverse/normal), and the **fixed** Marsden g(r) shape constants `ALN/NK/NM/NN/R0` as value-only columns. The constants select the force law; sources report them without uncertainties and they never appear in solution covariances, so they are not covariance dimensions. Null constants mean the asteroid convention g(r) = (1 au/r)².
- `SolvedStateCovariances` is deleted. `CoordinateCovariances` rows now hold either the familiar 6×6 or a 9×9 over the fixed basis (coordinates, A1, A2, A3), including cross-covariances. `to_matrix()` still returns the 6×6 block (existing consumers unaffected); `to_full_matrix()` / `nongrav_block_mask()` expose the extension. Coordinate transforms carry the full matrix (coordinate block via the Jacobian, non-grav dimensions via identity), `propagate_2body` carries it through the STM as dynamically inert, and `VariantOrbits` jointly samples the 9-dimensional state (whitened, so ~1e-26-scale A-variances survive).

### Importers and limitations

- **SBDB** (`query_sbdb_new`): builds the 9×9 in the native cometary basis; fills unlisted g(r) constants with the classic Marsden comet standard on any solution with fitted accelerations (SBDB lists constants only when they differ from it).
- **NEOCC**: Yarkovsky (`AMRAT`/`A2`) decode only; the AMRAT dimension is marginalized out with a warning.
- **Dropped with warnings, by policy**: `DT` (asymmetric outgassing — DT-fit comets like 67P/81P become symmetric approximations with km-scale model error through perihelion; no JPL-parity claims for those), `AMRAT`/`RHO` thermophysical models, and estimated shape parameters. Nothing is inferred silently.
- Near-parabolic element imports (e ≈ 1, e.g. C/2022 E3) hit a **pre-existing** cometary→Cartesian conversion defect, tracked separately in #210. Force-model parity claims are bounded to Cartesian-state propagation.

### Evidence

Companion `adam-assist` regressions propagate from JPL Horizons states with the cited SBDB solutions and match Horizons to 15.6 m (99942, inverse-square), 9.2 m (C/2022 E3, comet standard), and 179 m (1I/'Oumuamua, custom tuple) over ±300 days, against gravity-only controls of 3.7 km / 2,775 km / 2.8M km.

### Coordinated release order

1. Merge and release `adam_core` (this PR).
2. `adam-assist` (B612-Asteroid-Institute/adam-assist#34) and `mpcq` (B612-Asteroid-Institute/mpcq#21) then replace their moving `codex/nongrav-support` git dependencies with the tagged minimum version and run clean-install paired tests.

### Scope note

The branch also carries close-approach/impact-visualization changes that predate this review cycle; reviewers flagged them as candidates for splitting — author's call.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


akoumjian:
## Review: non-gravitational support

Reviewed locally (worktree off `codex/nongrav-support`). CI is green and a targeted local run was clean (`133 passed, 4 skipped`), so this is about **design coherence and consumption ergonomics**, plus a couple of correctness edges. The core abstraction — a labeled, dimension-generic solved-state covariance that transforms an orbital block via the Jacobian and carries extra parameters through an identity block — is genuinely elegant and worth keeping. Most of the notes below are about reducing *redundant sources of truth* and *stringly-typed* fields before this API ossifies.

---

### Correctness edges worth addressing before merge

**1. Impact visualization drops non-grav from the variant cloud.**
In `generate_impact_visualization_data`, `VariantOrbits` are rebuilt as `Orbits` with only `orbit_id`/`object_id`/`coordinates`:

```python
Orbits.from_kwargs(
    orbit_id=variant_orbits.variant_id,
    object_id=variant_orbits.object_id,
    coordinates=variant_orbits.coordinates,
)
```

A non-grav-aware propagator then integrates the cloud gravity-only while the nominal orbit may include non-grav forces — exactly the inconsistency this PR is trying to remove. Fix: carry `physical_parameters`, `non_gravitational_parameters` (and likely `solved_state_covariance`) through, or operate on the `VariantOrbits` directly.

**2. Zero-mean-but-uncertain non-grav slips past the unsupported-propagator warning.**
`has_non_gravitational_parameters()` treats exact-zero values as absent (fine for the propagation short-circuit), but with `covariance=True` a solved-state covariance can sample non-zero `A2` around a zero mean. A propagator without `supports_non_gravitational_forces` then silently ignores that uncertainty with no warning. Suggest the warning/covariance path also trigger on an estimated solved-state covariance (`dimension > 6`) or non-empty `estimated_parameter_names`, not only on non-zero nominal values. (See also #10 — splitting the predicate fixes both.)

**3. `VariantOrbits.collapse()` rebuilds the solved-state covariance only if the propagated *nominal* still carries one.** If a backend correctly nulls the epoch-specific solved covariance on output (as `propagate_2body` now does), `collapse()` won't rebuild it from the variants even though the variant cloud has everything needed. Consider keying the rebuild off the *variants'* parameter names rather than the nominal orbit's column.

---

### Data model & API coherence (the main thrust)

**4. The “what was estimated + its uncertainty” fact is stored 3–4 times and hand-synced.**
For a fitted `A2` the sigma/identity lives in `non_gravitational_parameters.A2_sigma`, in `estimated_parameter_names`, in the `solved_state_covariance` diagonal (whose `parameter_names[6:]` also names it), and the counts appear as `solution_dimension` / `parameter_count` / `solved_state_covariance.dimension`. The importers go to real lengths to keep these consistent, and the docs concede “downstream code that modifies one is responsible for the other” — the tell that there’s no single source of truth.
*Fix:* make the solved-state covariance authoritative for fitted uncertainties and derive `*_sigma` on access (e.g. a `sigmas_for(names)` helper), rather than materializing both. Keep `*_sigma` columns only for sources that report a sigma without a covariance.

**5. `coordinates.covariance` (6×6) and `solved_state_covariance[:6,:6]` are the same matrix by construction.** Both importers build the 6×6 and the leading block in the same basis, so they’re redundant at ingest and can silently diverge later.
*Fix (pick one):* store only the solved-state covariance and expose the 6×6 via the existing `to_orbital_covariances()`; **or** keep `coordinates.covariance` canonical and have `SolvedStateCovariances` store only the *extra* rows/cross-covariances, reconstructing the full matrix on demand.

**6. Two covariance containers with divergent return contracts and no labeled accessor.** `CoordinateCovariances.to_matrix()` → dense `(N,6,6)`; `SolvedStateCovariances.to_matrix()` → `list[np.ndarray | None]`. To read a solved matrix meaningfully you must `zip(to_matrix(), parameter_names_list())` by hand.
*Fix:* add a labeled accessor (e.g. `to_labeled()` returning name→index mapping or a small DataFrame per row), and align the two `to_matrix` signatures where possible.

**7. Structured label lists are encoded as CSV-in-a-cell.** `estimated_parameter_names` and `parameter_names` are comma-joined strings parsed back via `parse_parameter_names(...split(","))`. In an otherwise typed Arrow/quivr schema, a `large_list<string>` column is the coherent choice — no re-parsing at each call site, no implicit “names can’t contain commas” rule, and it’s introspectable.

**8. `model` is a free-text, source-specific taxonomy on the table billed as “canonical.”** Observed values: `"nongrav"`, `"yarkovsky"`, `"cometary"`, `"srp"`, `"neocc-model-2"`. The same physics gets different labels by provider, and `"neocc-model-N"` leaks the source’s internal code into the schema, so nothing downstream can reliably switch on it.
*Fix:* a controlled vocabulary / enum for `model`, with a separate raw-source field (`source_model_code`) for provenance.

**9. The canonical parameter list is duplicated.** The same 11 names live as `NON_GRAVITATIONAL_VALUE_FIELDS` (`non_gravitational_parameters.py`) and `NON_GRAVITATIONAL_PARAMETER_NAMES` (`solved_state_covariances.py`); adding a parameter means touching both tuples plus `nulls()` plus `_empty_nongrav_row()`. Consolidate to one definition imported where needed.

**10. `has_non_gravitational_parameters()` overloads its name.** A fully populated row with fitted `A2 = 0.0` answers `False`. That conflates “has force-relevant values” with “has a non-grav solution.”
*Fix:* split into `has_nonzero_forces()` (the propagation short-circuit) and a schema predicate like `is_estimated()` / `has_solution()`. This also resolves #2 cleanly.

**11. The transform API is asymmetric — and still has the variadic-arity pattern the PR set out to remove.** This PR rightly made `Orbits.to_keplerian()` always return bare coordinates. But:
- `coordinates_to(Keplerian)` returns coords and silently skips the solved-state covariance;
- `solved_state_covariance_to(Keplerian)` returns only the covariance (and re-runs the full transform, duplicating the Jacobian);
- there’s no single call returning a fully consistent transformed `Orbits` (coords + 6×6 + solved-state, labels in sync);
- `transform_coordinates(..., solved_state_covariances=...)` still returns *tuple-or-bare* depending on the kwarg — the exact flag-dependent arity removed elsewhere.
*Fix:* an `Orbits.to(representation, frame=..., origin=...) -> Orbits` that transforms coordinates and both covariances together in one Jacobian pass; keep the low-level `transform_coordinates` arity stable (always return the pair, or a small result object).

**12. Awkward toggle:** `with_non_gravitational_parameters(enabled=False)` is identical to `without_non_gravitational_parameters()`, and `with_X()` reads like it should add something it can’t. The `include_nongrav` flags plus `without_*` already cover the need — consider dropping `with_*`.

**13. Naming:** the column `solved_state_covariance` (singular) holds a `SolvedStateCovariances` (plural, N rows), unlike `coordinates` / `physical_parameters`. Minor, but visible at the schema level.

---

### Minor / polish

- `orbits/__init__.__all__` lists `VariantEphemeris`, but it’s never imported, so `from adam_core.orbits import *` raises `AttributeError`. Pre-existing on `main`, but this PR edits that exact `__all__` block to add the two new exports — a free fix while you’re there (`VariantEphemeris` lives in `.variants`).
- The PR body is empty for a ~5k-line, multi-source feature; a short summary of the new public surface (`NonGravitationalParameters`, `SolvedStateCovariances`, `include_nongrav`, the transform entry points) and a migration note for the `to_keplerian`/`solved_state_covariance_to` split would help reviewers and downstream users.

---

### What’s great

- Dimension-generic solved-state covariance with self-describing basis labels and identity-block transport of extra parameters.
- `include_nongrav` is uniform across `query_*`, `propagate_*`, `generate_ephemeris`, and `VariantOrbits.create`.
- Whitened-basis sampling so ~1e-26 non-grav variances aren’t truncated to zero spread.
- Tests assert values (sigma magnitudes, cross-covariance recovery), not just shapes.

None of 4–13 are correctness blockers, but they’re the difference between “works” and “pleasant to consume.” I’d prioritize **#4/#5 (one source of truth for uncertainty)**, **#7/#8 (type the labels and the model vocabulary)**, and **#11 (one symmetric transform)** before the API hardens, and fix **#1** regardless.

src/adam_core/orbits/query/neocc.py
```diff
@@ -210,10 +270,124 @@ def _physical_parameters_from_neocc(data: Dict[str, Any]) -> PhysicalParameters:
...
+    solve_for = _solve_for_codes_to_names(codes)
+    if len(solve_for) != len(codes):
+        # Unknown solve-for codes were dropped by the mapping.
+        return False
+    dimension = info.get("dimension") or 6
+    if dimension - 6 != len(solve_for):
+        return False
+    if info.get("model_used") not in (None, 0, 1):
+        return False
+    return all(name in _NEOCC_SUPPORTED_SOLVE_FOR for name in solve_for)
```
akoumjian, line 308: `a2 = vector[1]` in `_non_gravitational_parameters_from_neocc` assumes the NGR record is always the documented pair `(AMRAT, A2)`. If an OEF variant ever emits only the solved parameters (e.g. a lone A2 for a dimension-7 solution), we'd silently store a **null value against a real A2 variance** in the covariance block.

Cheap guard — any other layout degrades to the existing "unsupported model or solve-for parameters" warning path, leaving value and block null consistently:

```suggestion
    if info.get("model_used") not in (None, 0, 1):
        return False
    vector = info.get("vector") or []
    if vector and len(vector) != 2:
        # The Yarkovsky-model NGR record is the documented pair
        # (AMRAT [m^2/t], A2 [1e-10 au/d^2]); any other layout would make
        # the positional vector[1] read below decode the wrong value.
        return False
    return all(name in _NEOCC_SUPPORTED_SOLVE_FOR for name in solve_for)
```

src/adam_core/dynamics/tests/test_propagation.py
```diff
@@ -12,6 +12,7 @@
 from ...coordinates.origin import Origin
```
akoumjian, line 13: Needed for the extended-covariance test suggested at the end of the non-grav test block below.

```suggestion
from ...coordinates.covariances import CoordinateCovariances
from ...coordinates.origin import Origin
```

src/adam_core/dynamics/tests/test_propagation.py
```diff
@@ -505,6 +506,103 @@ def test_propagate_2body_preserves_physical_parameters():
...
+    )
+
+    propagated = propagate_2body(
+        orbits,
+        Timestamp.from_mjd([60000.0, 60001.0], scale="tdb"),
+        include_nongrav=False,
+    )
+
+    assert propagated.non_gravitational_parameters.A2[0].as_py() is None
+    assert not propagated.coordinates.covariance.has_nongrav_block()
```
akoumjian, line 604: Zero-mean-with-block is the *only* legal path for a 9×9 covariance through `propagate_2body`, and the "dynamically inert" docstring claim is currently untested end-to-end (the tests above cover reject / zero-value / strip, but none carries a covariance).

The construction below builds the 9×9 as `M P Mᵀ` with a scalar latent `w` driving A1/A2 at a fixed 1:2 ratio — PSD by construction, and it gives a Jacobian-free oracle: any *linear* transport must preserve the 1:2 ratio between the propagated cross-covariance columns (the factor 2 is a power of two, so the equality is exact even in floats).

```suggestion
    assert not propagated.coordinates.covariance.has_nongrav_block()


def test_propagate_2body_carries_extended_covariance_through_stm():
    # Zero-mean A1/A2 with real variance and state cross-covariances is the
    # one legal path for a 9x9 covariance through 2-body dynamics: the
    # coordinate block gets the STM, the cross-covariances rotate with it,
    # and the non-gravitational block is dynamically inert.
    cov6 = np.diag([1e-14, 1e-14, 1e-14, 1e-16, 1e-16, 1e-16])
    P = np.zeros((7, 7))
    P[:6, :6] = cov6
    P[6, 6] = 1.0
    P[0, 6] = P[6, 0] = 0.3 * np.sqrt(cov6[0, 0])  # corr(x, w) = 0.3

    a = 1e-13  # au/d^2 scale of the A-parameter uncertainty
    M = np.zeros((9, 7))
    M[:6, :6] = np.eye(6)
    M[6, 6] = a  # A1 = a * w
    M[7, 6] = 2.0 * a  # A2 = 2a * w
    full = M @ P @ M.T

    def _orbit(covariance: np.ndarray) -> Orbits:
        return Orbits.from_kwargs(
            orbit_id=["o1"],
            object_id=["o1"],
            non_gravitational_parameters=NonGravitationalParameters.from_kwargs(
                source=["SBDB"],
                A1=[0.0],
                A2=[0.0],
                A3=[None],
            ),
            coordinates=CartesianCoordinates.from_kwargs(
                x=[1.0],
                y=[0.0],
                z=[0.0],
                vx=[0.0],
                vy=[0.017],
                vz=[0.0],
                covariance=CoordinateCovariances.from_matrix(
                    covariance[np.newaxis, ...]
                ),
                time=Timestamp.from_mjd([60000.0], scale="tdb"),
                origin=Origin.from_kwargs(code=["SUN"]),
                frame="ecliptic",
            ),
        )

    times = Timestamp.from_mjd([60030.0], scale="tdb")
    propagated = propagate_2body(_orbit(full), times)
    control = propagate_2body(_orbit(full[:6, :6].copy()), times)

    out = propagated.coordinates.covariance
    assert out.nongrav_block_mask().tolist() == [True]
    full_out = out.to_full_matrix()[0]

    # The A-block is dynamically inert under 2-body dynamics.
    np.testing.assert_array_equal(full_out[6:, 6:], full[6:, 6:])

    # The coordinate block matches a plain 6x6 propagation of the same orbit.
    np.testing.assert_allclose(
        full_out[:6, :6],
        control.coordinates.covariance.to_matrix()[0],
        rtol=1e-12,
    )

    # The cross-covariances were transformed (not copied verbatim) ...
    assert not np.allclose(full_out[:6, 6], full[:6, 6])
    # ... by a linear map: the exact 1:2 column ratio survives the STM.
    np.testing.assert_allclose(full_out[:6, 7], 2.0 * full_out[:6, 6], rtol=1e-12)
```

src/adam_core/coordinates/covariances.py
```diff
@@ -100,33 +131,102 @@ def to_matrix(self) -> np.ndarray:
...
+            if value.size == COORD_DIM * COORD_DIM:
+                full[i, :COORD_DIM, :COORD_DIM] = value.reshape(COORD_DIM, COORD_DIM)
+            elif value.size == FULL_DIM * FULL_DIM:
+                full[i] = value.reshape(FULL_DIM, FULL_DIM)
+            else:
+                raise ValueError(
+                    f"Covariance row {i} has {value.size} values; expected "
+                    f"{COORD_DIM * COORD_DIM} or {FULL_DIM * FULL_DIM}."
+                )
+        return full
```
akoumjian: Per-row Python loop in a path that's hotter than it looks: `to_matrix()` delegates here whenever **any** row carries a block, so a single non-grav orbit drags every `to_matrix()` call on the table (residuals, ephemeris, OD) through this loop, and `from_matrix` (below) runs on every `propagate_2body` output.

Row lengths only take two values, so this collapses to two vectorized gathers (~50–100× on large tables); the existing round-trip and mixed-row tests cover the change:

```suggestion
        arr = self.values
        if isinstance(arr, pa.ChunkedArray):
            arr = arr.combine_chunks()
        flat = arr.values.to_numpy(zero_copy_only=False)
        offsets = arr.offsets.to_numpy(zero_copy_only=False)
        starts = offsets[:-1]
        lengths = np.diff(offsets)
        valid = ~arr.is_null().to_numpy(zero_copy_only=False)

        is6 = valid & (lengths == COORD_DIM * COORD_DIM)
        is9 = valid & (lengths == FULL_DIM * FULL_DIM)
        bad = valid & ~is6 & ~is9
        if bad.any():
            i = int(np.flatnonzero(bad)[0])
            raise ValueError(
                f"Covariance row {i} has {lengths[i]} values; expected "
                f"{COORD_DIM * COORD_DIM} or {FULL_DIM * FULL_DIM}."
            )
        if is6.any():
            gather = starts[is6, None] + np.arange(COORD_DIM * COORD_DIM)
            full[is6, :COORD_DIM, :COORD_DIM] = flat[gather].reshape(
                -1, COORD_DIM, COORD_DIM
            )
        if is9.any():
            gather = starts[is9, None] + np.arange(FULL_DIM * FULL_DIM)
            full[is9] = flat[gather].reshape(-1, FULL_DIM, FULL_DIM)
        return full
```

src/adam_core/coordinates/covariances.py
```diff
@@ -100,33 +131,102 @@ def to_matrix(self) -> np.ndarray:
...
+            else:
+                rows.append(np.asarray(cov, dtype=np.float64).reshape(-1))
+            offsets.append(offsets[-1] + rows[-1].size)
+        flat = np.concatenate(rows) if rows else np.empty(0, dtype=np.float64)
+        return cls.from_kwargs(
+            values=pa.LargeListArray.from_arrays(
+                pa.array(offsets, type=pa.int64()),
+                pa.array(flat, type=pa.float64()),
+            )
+        )
```
akoumjian: Same pattern for the write side — two vectorized scatters instead of the per-row loop:

```suggestion
        is6 = np.isnan(covariances[:, COORD_DIM:, :]).all(axis=(1, 2)) & np.isnan(
            covariances[:, :, COORD_DIM:]
        ).all(axis=(1, 2))
        lengths = np.where(
            is6, COORD_DIM * COORD_DIM, FULL_DIM * FULL_DIM
        ).astype(np.int64)
        offsets = np.zeros(len(covariances) + 1, dtype=np.int64)
        np.cumsum(lengths, out=offsets[1:])
        flat = np.empty(offsets[-1], dtype=np.float64)
        if is6.any():
            scatter = offsets[:-1][is6, None] + np.arange(COORD_DIM * COORD_DIM)
            flat[scatter] = covariances[is6][:, :COORD_DIM, :COORD_DIM].reshape(
                scatter.shape[0], -1
            )
        if (~is6).any():
            scatter = offsets[:-1][~is6, None] + np.arange(FULL_DIM * FULL_DIM)
            flat[scatter] = covariances[~is6].reshape(scatter.shape[0], -1)
        return cls.from_kwargs(
            values=pa.LargeListArray.from_arrays(
                pa.array(offsets, type=pa.int64()),
                pa.array(flat, type=pa.float64()),
            )
        )
```

src/adam_core/coordinates/covariances.py
```diff
@@ -37,31 +44,55 @@ def sigmas_to_covariances(sigmas: np.ndarray) -> np.ndarray:
...
 
     values = qv.LargeListColumn(pa.float64(), nullable=True)
-    # When fixed, we should revert to:
-    # values = Column(pa.fixed_shape_tensor(pa.float64(), (6, 6)))
 
     @property
     def sigmas(self):
         cov_diag = np.diagonal(self.to_matrix(), axis1=1, axis2=2)
         sigmas = np.sqrt(cov_diag)
         return sigmas
```
akoumjian, line 64: Worth a note now that extended rows exist — `sigmas` (like `to_matrix`) reports the coordinate block only:

```suggestion
        # NOTE: coordinate (6x6) block only; the non-gravitational
        # (A1, A2, A3) dimensions of extended rows are not included --
        # take the diagonal of `to_full_matrix` for those.
        sigmas = np.sqrt(cov_diag)
        return sigmas
```

src/adam_core/orbits/query/sbdb.py
```diff
@@ -536,6 +655,28 @@ def _orbits_from_sbdb_payloads(
             if data.ndim != 2 or data.shape[0] < 6 or data.shape[1] < 6:
                 raise ValueError("Expected SBDB covariance matrix to be at least 6x6.")
             cov_matrix = data[:6, :6]
+            if data.shape[0] > 6:
```
akoumjian: Not a one-click suggestion because the label check sits just above this hunk (it's pre-existing on main), but this PR leans on it much harder now: non-grav objects are exactly the ones whose covariances carry >6 labels, and a single object with an unexpected label layout raises out of `_orbits_from_sbdb_payloads` and fails the **entire** `query_sbdb_new` batch.

There's already a graceful path right below (the per-element sigma fallback when `cov_matrix is None`) — suggest routing label surprises into it:

```python
        cov = orbit.get("covariance")
        cov_matrix: np.ndarray | None = None
        labels_ok = True
        if isinstance(cov, dict) and cov.get("data") is not None:
            labels = cov.get("labels")
            if isinstance(labels, list):
                labels6 = [str(x) for x in labels[:6]]
                if labels6 != expected_labels:
                    logger.warning(
                        "SBDB covariance labels for object %s start with %s, "
                        "expected %s; discarding the covariance and falling "
                        "back to per-element sigmas.",
                        obj_id,
                        labels6,
                        expected_labels,
                    )
                    labels_ok = False

            if labels_ok:
                data = np.asarray(cov["data"], dtype=np.float64)
                # ... existing body unchanged (6x6 slice, 9x9 conversion,
                # cov["elements"] preference), indented one level ...
```

Deliberate side effect: when the labels can't be interpreted we also skip `cov["elements"]`/`cov["epoch"]` — if the label alignment can't be trusted, the element-vector alignment shouldn't be either. The validation inside `_convert_sbdb_full_covariance` can stay as defense for other callers.

akoumjian (changes requested):
couple small changes

src/adam_core/orbits/query/sbdb.py
KatKiker: Updated. Included the dimension check so a single malformed covariance doesn't fail the whole query for either reason. Added test: test_query_sbdb_unexpected_covariance_labels

akoumjian:
Follow-up after rechecking the latest head (`3ded4b9d`): the new commits correctly fix the scale-blind STM cross-covariance assertion and update the MPCQ integration fixture to the documented raw `1e-10 au/d^2` contract. The focused non-grav suite now passes locally (`139 passed, 4 skipped`), and the corrected paired MPCQ integration test passes against MPCQ #21.

Remaining instructions before merge:

1. **Rebase onto current `main`.** Against `030cf9e9`, the sole merge conflict is still `src/adam_core/orbits/__init__.py`. Resolve it while retaining imports/exports for all of `Trajectory`, `NonGravitationalParameters`, `VariantEphemeris`, and `VariantOrbits`.
2. **Run complete CI on the rebased/merged result.** GitHub still reports this PR as `CONFLICTING` / `DIRTY` / `CHANGES_REQUESTED`, and currently exposes only the documentation check. The focused branch tests are now clean, but a current merged-head full test run and normal re-review are still required.
3. **Keep hyperbolic conversion claims bounded.** Fix the existing cometary-to-Cartesian defect or track it separately and document it before claiming complete SBDB-state-to-ASSIST parity for hyperbolic objects such as C/2022 E3 or 1I. The Horizons-anchored assist regressions intentionally isolate force integration from that conversion path.
4. **Document the PR and release sequence.** This PR still has an empty body. Summarize the canonical A1/A2/A3 plus fixed-shape model, 9x9 covariance behavior, importer limitations, and the coordinated release order. Release core first so adam-assist and MPCQ can replace their moving branch dependencies with a tagged minimum version and run clean-install paired tests.

Native `DT` support is not required for merge under the agreed policy: dropping it with a clear warning and applying the symmetric approximation is acceptable, without making stronger JPL-parity claims for DT-bearing solutions. Likewise, unsupported `AMRAT`/`RHO` and estimated-shape cases should continue to fail/degrade explicitly rather than be inferred silently.

PR #206, opened 2026-07-15
Claim period-doubled folds when doubling is the sole remaining ambiguity

ntellis:
## What

When the best fit is a single-peaked fold that the solver has already doubled on the two-peaks-per-rotation shape prior, and no rival alias cluster survives the 95% competitiveness cut, the verdict may now be `single_period`, with a new `doubled_fold_assumed` confidence flag disclosing the assumption. Previously such cases were always demoted to `period_family` with reason `single_max_alias`. A new parameter `claim_doubled_fold` (default `True`) restores the strictly agnostic behavior when `False`. The long-period and sub-harmonic guardrails still apply to these claims, and the period search itself is untouched (recovered periods are bit-identical; only the verdict and disclosure change).

## Why

A gate-attribution study on the validation sets asked why the solver claims so few periods on sparse survey cadence. The dominant binder (multiple surviving alias clusters) is not a tuning knob, but a clean secondary population emerged: objects whose *only* remaining ambiguity is the half/full octave of a doubled fold. For an elongated body, the doubled period is the physical answer; refusing to claim it costs correct claims without preventing any wrong ones on the measured sets.

## Measured effect (validation sets, offline counterfactual + live confirmation)

| set | claims before | claims after | strict precision | wrong-family |
|---|---:|---:|---|---|
| LCDB standard candles (115 solved) | 43 | 48 | 36/43 -> 41/48 | 1 (unchanged: 3295 Murakami) |
| DAMIT shape models (60) | 33 | 35 | 32/33 -> 34/35 | 1 (unchanged: 478 Tergeste) |
| Rubin survey regression (76) | 6 | 11 | 6/6 -> 11/11 | 0 |

Pooled calibration precision rises 0.895 -> 0.904; no new wrong-family claim on any set.

**All 12 predicted verdict flips were then confirmed live with this branch** (validated_staged + JAX, each source's stored knobs): 209 Dido, 829 Academia, 3055 Annapavlova, 3975 Verdi, 1323 Tugela (LCDB); 18 Melpomene, 896 Sphinx (DAMIT); 2025 MD76, MG56, MO35, MP71, MT24 (survey). Every one claims `single_period` with `doubled_fold_assumed` set, strictly within its tolerance (worst error 0.17%), and no guardrail intervened.

## Risk and disclosure

A genuinely single-peaked (spheroidal) body would make a doubled-fold claim a 2x alias — the cardinal failure mode. That is why the assumption is a *flag*, not silence: bulk consumers can filter `doubled_fold_assumed` and treat these claims separately, and `claim_doubled_fold=False` restores the pure hedge. On the measured sets the failure did not occur.

## Testing

- New behavioral test on the committed 1323 Tugela fixture (the canonical doubled-only case): default claims with the flag; `claim_doubled_fold=False` reproduces the previous `period_family` / `single_max_alias` contract with an identical period. Runs in ~11 s.
- Full rotation suite green under the new default: 29 passed, 1 xfailed — the zero-false-confidence CI gate still xfails on the known 1627 Ivar sparse-fixture alias, and Tugela's new claim is within tolerance, so the gate's semantics are unchanged.
- `mypy --strict` clean on the subsystem; ruff/black/isort clean; docs build clean under `-W` (cookbook section added documenting the assumption, the measured numbers, and the opt-out).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01CZNKbKYQR3hFiP9vLdzZU4

### DIFF
--- a/docs/source/cookbook/rotation_periods.rst
+++ b/docs/source/cookbook/rotation_periods.rst
@@ -55,6 +55,18 @@ Note that ``single_period`` is high-confidence but **not a zero-alias guarantee.
   ``single_period`` result against ``alternate_period_days`` and
   ``reliability_code`` rather than treating it as infallible.
 
+One disclosed assumption: when the best fit is a single-peaked fold whose only
+surviving ambiguity is the half/full octave, the solver doubles the period on
+the two-peaks-per-rotation shape prior and, by default, claims
+``single_period`` with the confidence flag ``doubled_fold_assumed``. Measured
+on the LCDB/DAMIT standard-candle sets this raises confident claims (76 to 83
+pooled) with strict precision improving and no new wrong-family claim; on the
+Rubin survey regression it raises claims from 6 to 11 of 76, all strictly
+correct. A genuinely single-peaked (spheroidal) body would make such a claim a
+2x alias, so pass ``claim_doubled_fold=False`` to keep these cases as
+``period_family`` (reason ``single_max_alias``). Filter on the flag to treat
+them separately in bulk results.
+
 Data Requirements
 -----------------
 
--- a/src/adam_core/photometry/rotation/estimator.py
+++ b/src/adam_core/photometry/rotation/estimator.py
@@ -982,6 +982,7 @@ def _classify_confidence(
     observation_count_sufficient: bool,
     is_reliable: bool,
     is_valid: bool,
+    claim_doubled_fold: bool,
 ) -> tuple[str, str, list[str], list[str]]:
     """Single deterministic verdict from the confidence classifier's decision tree.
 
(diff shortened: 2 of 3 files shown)


ntellis:
Full-pipeline revalidation on this branch (after merging main at a4250947), measured end to end rather than counterfactually:

| Set | Claims (before -> after) | Precision |
|---|---|---|
| LCDB 118, best apparition per object | 43 -> **51** | strict 43/51 (0.843, was 0.837); within tolerance of a harmonic 50/51; wrong-family: Murakami only (unchanged) |
| DAMIT 61 | 33 -> **35** | strict 34/35; wrong-family: Tergeste only (unchanged) |
| Rubin survey 76 | 6 -> **11** | 11/11 within 2% of the published period (worst 1.19%); 5 of 11 carry `doubled_fold_assumed` |

The LCDB gain exceeds the +5 predicted in the PR body because verdict flips happen per apparition *before* best-apparition selection, so newly-confident apparitions can also win the selection. Pooled LCDB+DAMIT strict precision is 77/86 = 0.895 (pre-change: 68/76 = 0.895) - claims up 13% at flat measured precision, and zero new wrong-family claims anywhere.

Fresh artifacts (calibration JSONs, per-apparition solves, survey diagnostics) are in the coordination workspace under `.scratch/blog/`, with the pre-change versions preserved in `.scratch/blog/pre_doubled_fold_backup/`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01XecpgSN9YoZ3NVfSNC6Bzk

